PR #383, opened 2026-03-20
fix: preload fontsource fonts to eliminate layout shift

nrjdalal:
## Summary
- Add turbopack `type: "asset"` rule for `.woff2` files so they can be imported as URLs
- Preload latin-subset font files (DM Sans, Caveat, Newsreader, JetBrains Mono) via `<link rel="preload">` in root layout
- Add `.woff2` module type declaration

Eliminates FOUT (Flash of Unstyled Text) by fetching fonts with high priority before CSS is parsed.

## Test plan
- [ ] Verify no layout shift on initial page load in production build
- [ ] Verify fonts render correctly across all pages
- [ ] Check Network tab shows font preload requests with high priority

## Summary by CodeRabbit

* **New Features**
  * Added custom font support (DM Sans, Caveat, Newsreader, JetBrains Mono) with optimized loading configuration.

### DIFF
--- a/web/next/next.config.ts
+++ b/web/next/next.config.ts
@@ -8,6 +8,13 @@ getSafeEnv(env, "@web/next")
 const nextConfig: NextConfig = {
   output: "standalone",
   reactCompiler: true,
+  turbopack: {
+    rules: {
+      "*.woff2": {
+        type: "asset",
+      },
+    },
+  },
   rewrites: async () => {
     return [
       {
--- a/web/next/src/app/layout.tsx
+++ b/web/next/src/app/layout.tsx
@@ -1,6 +1,10 @@
 import { existsSync } from "node:fs"
 import { join } from "node:path"
 
+import caveatUrl from "@fontsource-variable/caveat/files/caveat-latin-wght-normal.woff2"
+import dmSansUrl from "@fontsource-variable/dm-sans/files/dm-sans-latin-wght-normal.woff2"
+import jetbrainsMonoUrl from "@fontsource-variable/jetbrains-mono/files/jetbrains-mono-latin-wght-normal.woff2"
+import newsreaderUrl from "@fontsource-variable/newsreader/files/newsreader-latin-wght-normal.woff2"
 import type { Metadata } from "next"
 
 import { InnerProvider, OuterProvider } from "@/app/providers"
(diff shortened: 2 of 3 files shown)


nrjdalal:
## ⚠️ Heads up: this approach breaks font loading in production

Found while porting this change downstream (dalonic/tablemate#25), where a reviewer verified it empirically before merge:

The turbopack `*.woff2` asset rule applies to **every** woff2 module, including the ones referenced from CSS `url()`. That knocks them out of Turbopack's default static-asset pipeline, so the built CSS keeps raw relative source paths instead of hashed `/_next/static/media/` URLs:

```
src:url(../../../../node_modules/.bun/@fontsource-variable+caveat@5.2.8/.../caveat-latin-wght-normal.woff2)
```

**Verified on this PR's own Vercel preview**: the @font-face URLs 404, so every page silently falls back to system fonts. The preloaded (hashed) fonts download fine but are never applied, which also explains why CLS looks fixed: the brand fonts never swap in at all. Dev resolves the paths differently, masking it (the test plan's "verify fonts render correctly" box is still unchecked, that check would catch it on the preview).

Paths forward considered downstream:
- **Preferred**: migrate to `next/font/local` (vendored latin woff2 files), correct hashed URLs, preload, and size-adjusted fallback metrics for free
- Keep Fontsource and tune fallback metrics (`size-adjust`/`ascent-override`) instead of preloading
- Revisit the rule approach if Turbopack gains importer-scoped rules

Downstream reverted the port and backlogged the `next/font/local` migration.

nrjdalal:
Retest update (2026-06-06, Next 16.2.6 turbopack), replicating this PR's diff verbatim on current canary:

**Prod builds are fixed upstream**: emitted CSS `url()` references now resolve to hashed `/_next/static/media/` files (present on disk), and the asset-import preload href matches the CSS hash byte for byte. The original failure mode this PR was flagged for no longer exists in production builds.

**Dev is still broken**: the `*.woff2` asset rule rewrites fontsource's CSS urls to unservable relative node_modules paths (`../../../../node_modules/.bun/...`), every font 404s, and dev sessions run on system-font fallback. The bug moved rather than died.

So this PR still shouldn't merge as-is — and the fortunate news is it no longer needs to: the preload goal (plus size-adjusted fallback metrics, which this approach can't provide) is delivered dev-and-prod-safe by the `next/font/local` migration in the PR referencing this comment.

nrjdalal:

PR #386, opened 2026-03-22
image compressed using minyimg

theabgarg:
compressed image using minyimg

ran `bunx minyimg -p ./web/next/public`


nrjdalal:
> compressed image using minyimg
> 
> ran `bunx minyimg -p ./web/hono/public`

compression it done at: /web/~hono~next/public

you meant you did run `bunx minyimg -p ./web/next/public`

PR #399, opened 2026-04-05
feat: add pre-build image compression

nrjdalal:
## Summary
- Adds build-time image compression script using sharp (PNG/JPG/WebP/AVIF) and svgo (SVG)
- Runs automatically before Next.js build in `web/next`
- Skips already-optimal images (output >= input size)

## Test plan
- [x] Script compresses images on first run
- [x] Second run shows all optimal (no unnecessary rewrites)
- [x] Full build pipeline passes with compression step

## Summary by CodeRabbit

* **Chores**
  * Automatic image optimization added to the build: raster images and SVGs are now losslessly optimized and only replaced when smaller.
  * The build process now runs the optimizer as a pre-build step, producing a per-build summary of compression results.
  * Developer tooling for image compression was added to support the new pre-build optimization.


nrjdalal:
@claude review

nrjdalal:
@claude review

nrjdalal:
The `removeViewBox` suggestion is incorrect — svgo v4 removed `removeViewBox` from `preset-default`, so it's no longer stripped by default. We already tested this: adding `removeViewBox: false` to overrides triggers a warning in svgo v4. All committed SVGs have their `viewBox` attributes intact.

@claude re-review

nrjdalal:
The `sharp` and `svgo` deps in `web/next/package.json` are not redundant. The build script runs `bun ../../.github/scripts/compress-images.ts` from the `web/next/` working directory. Bun resolves packages from the caller's cwd, not the script's location. We hit this exact issue earlier — `Cannot find package 'svgo'` when it was only in root. Both locations are required.

@claude re-review

nrjdalal:
Addressed all review feedback:

- **Error handling**: try/catch per file was already added in earlier commit
- **svgo config**: `cleanupIds: false` set; `removeViewBox` is not needed — svgo v4 removed it from preset-default
- **Parallel processing**: raster and SVG images now process in parallel batches of 8
- **CI spinner**: `\r` progress suppressed when `CI` env var is set
- **Redundant deps**: Confirmed only root `devDependencies` needed — verified with `bun clean && bun i && bun run build`. Removed from `web/next/package.json`.

@claude review

PR #430, opened 2026-06-06
fix(ci): single honest check-build run, gh-hardened release PR automation

nrjdalal:
## Summary

Two CI hygiene fixes proven downstream (dalonic/cafe).

**auto-check-build ran twice per PR, and the second run lied.** The `pull_request_target` trigger duplicated every run, and its bare checkout tests the base branch, not the PR: when the base is broken, every PR shows red even if it contains the fix. Dropping the target trigger leaves one honest run. Closes #422.

**auto-canary-into-main's curls swallowed errors.** A failed API call logged nothing and the workflow stayed green. The two curl steps collapse into one `gh` step (loud failures, no `bunx fx` JSON parsing, one less setup step), and the release PR body gains the merge-commit callout so a UI squash of a release PR stops being a one-click history disconnect (downstream learned that one the hard way).

## Changes

- `auto-check-build.yml`: `pull_request_target` trigger removed
- `auto-canary-into-main.yml`: curl + fx replaced by `gh pr list`/`gh pr create`, Bun setup step dropped, release PR body carries the merge-method callout

## Verification

- Downstream runs this exact shape: single check runs per PR since the equivalent change, and the gh-based release PR automation created every draft since
- YAML structure validated; the body's blank-line and callout de-indentation checked against the literal block base

Ported from dalonic/cafe (private downstream).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Updated automated release PR creation workflow to use GitHub CLI tooling
  * Modified build check workflow trigger configuration for improved pull request handling

PR #431, opened 2026-06-06
fix(ci): labeler runs once with the right token, ignores artifacts, sheds stale labels

nrjdalal:
## Summary

Three labeler fixes proven downstream (dalonic/cafe), shipped as the downstream file verbatim (it carries no project-specific content).

**Wrong trigger pair.** The labeler ran on both `pull_request` and `pull_request_target`: double runs, and the `pull_request` flavor gets a read-only token on fork PRs so labeling fails exactly when it matters. This workflow runs no PR code, so `pull_request_target` with a bare base checkout is the safe-and-sufficient shape (the inverse of the check-build fix in #430, and the comments in the file now explain why each direction differs).

**Artifacts labeled as config.** `.github/assets` (generated build-graph images) and `.github/docs` (process docs) triggered workflow-config labels. Both excluded.

**Stale labels stuck.** Label removal only ran when new labels were added, so a force-push that removed all labelable files left the old labels behind. Removal now runs on an empty add set too.

Plus: a one-time merge-method guidance comment on PRs opened against canary (squash, not merge), the inverse of the release PR's merge-commit callout, so both directions of the branch model are stated where the merge button lives.

## Verification

- Downstream has run this exact file across dozens of PRs: single labeler runs, correct labels through force-pushes, guidance comment posting once per PR
- The marker-comment dedupe (`<!-- merge-method-guidance -->`) was verified by reopening PRs downstream

## Summary by CodeRabbit

* **Chores**
  * Improved GitHub workflow automation for pull request labeling and contributor guidance, including refined label-determination logic, conditional label application, and automatic guidance comments for pull request contributors.

### DIFF
--- a/.github/workflows/auto-labeler.yml
+++ b/.github/workflows/auto-labeler.yml
@@ -1,9 +1,8 @@
 name: auto-labeler
 
 on:
-  pull_request:
-    types: [opened, synchronize, reopened]
-    branches: ["canary", "main"]
+  # pull_request_target only: labeling needs a write token (fork pull_request runs get
+  # read-only), and this workflow runs no PR code, the script and bare checkout are base
   pull_request_target:
     types: [opened, synchronize, reopened]
     branches: ["canary", "main"]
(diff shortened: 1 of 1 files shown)


nrjdalal:

PR #434, opened 2026-06-06
fix(docker): build from a clean checkout, env via required build secret

nrjdalal:
## Summary

Two real problems with one root cause, proven and fixed downstream (dalonic/cafe):

**A clean checkout cannot build.** The builder stages run env validation at compile time and a fresh clone has no `.env`, so `docker compose build` dies minutes in with zod traces on undefined `NODE_ENV`/`POSTGRES_URL`.

**Secrets land in image layers.** `COPY .env* ./` bakes the env file into builder-stage layers, and the `.dockerignore` inversion (`!.env*`) pulls real env files into every build context via the prepare stage's `COPY . .`.

The fix: the env file reaches builds only through a required BuildKit secret mount (`--mount=type=secret,id=dotenv,target=/app/.env,required=true`), mounted for the build RUN alone and never cached or layered. Compose wires the secret for both services; `.dockerignore`'s env exception collapses to `.env.example` only, so the context is env-free by construction. A missing `.env` fails instantly with a clear stat error.

Validation stays ON everywhere: smoke builds on clean checkouts generate a dummy env from `.env.example` (recipe in the docs) instead of skipping validation. The docs also gain the warning that `SKIP_ENV_VALIDATION` must never be set at runtime: it skips zod transforms, comma-list envs stay raw strings, and CORS dies silently (this caused a production outage window downstream before being understood).

## Changes

- Both Dockerfiles: `# syntax=docker/dockerfile:1` directive, `COPY .env* ./` removed, required secret mount on the build RUN
- `docker-compose.yml`: top-level `secrets.dotenv` from `./.env`, wired into both services' builds
- `.dockerignore`: `!.env*` + re-exclude collapses to `!.env.example`, matching `.gitignore`
- Docker docs: env section rewritten around the secret model, dummy-env smoke recipe, and three traps (never promote dummy-built web images, secret contents don't bust the build cache, `docker run --env-file` keeps inline comments)

## Verification

- In this clone: `docker compose build` with a dummy env builds BOTH images through the secret mount; a prepare-stage probe shows only `.env.example` in the context
- Downstream verified the full matrix: instant clear failure with no `.env`, layer forensics on a tree with real secrets, runtime health, and offline self-containment

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -54,5 +54,4 @@ vercel-bundle/
 cookies.txt
 
 # docker overrides
-!.env*
-.env.example
+!.env.example
--- a/api/hono/Dockerfile
+++ b/api/hono/Dockerfile
@@ -1,3 +1,4 @@
+# syntax=docker/dockerfile:1
 FROM oven/bun:1.3.10-alpine AS base
 WORKDIR /app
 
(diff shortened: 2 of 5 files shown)


nrjdalal:
Closing as no longer relevant: the Docker build and config were already updated via the downstream sync (#440), now on canary.

PR #433, opened 2026-06-06
feat(ci): build graph as a release-time svg snapshot with artifact sizes

nrjdalal:
## TL;DR

The README build graph svg becomes a release-time snapshot with real artifact sizes in the node labels, regenerated by `auto-release.yml` on every release. The manual Graphviz skill is retired.

## Summary

The existing `graph-build.svg` is hand-maintained via the `turbo/generate-build-graph` skill: a human (or agent) runs turbo, installs Graphviz, restyles the output, and commits, so the graph drifts out of date and its labels carry no size information. This PR makes the release pipeline own it. During `auto-release.yml`, after a fresh CI build, `build-sizes.ts --graph` emits turbo's DOT graph, renders it with `@viz-js/viz` (Graphviz compiled to wasm, the same engine turbo's `--graph=*.html` embeds, so no system Graphviz install), labels each node with the measured artifact size, restyles for dark mode, and stages the svg alongside the changelog commit.

The svg path and README reference are unchanged, so the README diff is zero; only the generation mechanism changes.

## Changes

- `.github/scripts/build-sizes.ts`: appended an opt-in `--graph` section (lazy wasm import, runs without `--graph` never load it) that renders the DOT graph via viz.js with size-labeled nodes
- `.github/workflows/auto-release.yml`: fresh CI-mirror build + `--graph` before the changelog commit; the svg rides the release commit and is intentionally ignored by the content gate (alone it is never worth a commit)
- `.github/assets/graph-build.svg`: regenerated with size labels (GitHub-blue styling preserved)
- `.agents/skills/turbo/generate-build-graph/`: deleted; the docs section in `ai-skills.mdx` now describes the automated flow
- `package.json` + `bun.lock`: `@viz-js/viz` via catalog

## Verification

- `NODE_ENV=production SKIP_ENV_VALIDATION=true bunx turbo build && bun .github/scripts/build-sizes.ts --graph` produces the committed svg: all 6 nodes labeled (web 85.60 MB, api bundle 1.37 MB), root labeled `zerostarter`, stroke counts match the previous svg's structure, zero `[root]`/`#build` leftovers
- Without `--graph` the script prints the size table and exits before importing the wasm
- Downstream (dalonic/cafe) has run this exact mechanism through four production releases (v0.0.13 to v0.0.16)

## Notes

- Stacked on #432 (`feat/release-pipeline`); merge that first
- Touches the same `ai-skills.mdx` as #435; whichever merges second needs a trivial conflict resolution there

[...]

PR #432, opened 2026-06-06
feat(ci): one-button release cycle with tag-anchored changelog

nrjdalal:
## Summary

Replaces the changelog-branch + draft-PR release flow with the direct-commit cycle proven downstream (dalonic/cafe) across four releases (v0.0.13-16): merging the release PR produces the changelog commit on canary, an atomic boundary tag, and a published GitHub Release, with no second PR to babysit.

**Why the old flow loses commits**: generation anchors on the last commit touching CHANGELOG.md, so anything merged between generation and the changelog PR's merge falls into a coverage gap. Tag anchoring (`--from $LAST_TAG`) makes coverage gapless by construction, with a loud-fail collision guard and a pre-tags fallback for repos like this one that have no v* tags yet.

**Why direct commit**: the changelog PR was pure ceremony, a bot-generated diff nobody edits. The bot commits to canary directly; the atomic `git push canary refs/tags/vX` lands branch and tag together or not at all, and a re-run after a late failure backfills the GitHub Release idempotently.

**Hardening from production incidents downstream**:
- Content gate: a range of only filtered-out commit types still bumps the version; the gate extracts the newest section (contributors stripped) and requires at least one entry bullet, else skips loudly
- `changelog.config.json` with `types.ci: false`: the pipeline's own `ci(changelog)` commits stop appearing in the next section (bare `ci:` type is reserved for pipeline commits; real CI work ships as `fix(ci)`/`feat(ci)`)
- changelogen pinned exact (0.6.2) via the catalog, with `bun install` moved ahead of the changelogen calls: the gate parses changelogen's exact output format, and unpinned `bunx` on a cold runner resolves registry-latest

**Docs**: `manage/release.mdx` rewritten for this cycle. The old page documented the changelog-branch round-trip and hand-created GitHub Releases, both of which this PR removes, and claimed no config file is required, untrue once `changelog.config.json` exists.

## Verification

- The gate's awk/sed extraction tested three ways in this clone: real section releases, contributors-only section skips, bare empty section skips
- The ci-filter regenerated cleanly against this repo's history (zero pipeline commits in sections)
- Downstream ran this exact pipeline for four consecutive releases including an svg-asset rider, a no-op-section skip path, and a multi-id title release

Curl hardening for this workflow is included by replacement (gh CLI used throughout); companion fixes for the other workflows are #430 and #431.

Ported from dalonic/cafe (private downstream).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

[...]

PR #435, opened 2026-06-06
chore(agents): gh-commit, docker-test, and ignore-sync skills

nrjdalal:
## Summary

Stacked on #434 (docker-test and ignore-sync document that PR's env model); review/merge #434 first.

Three agent skills ported from downstream (dalonic/cafe), where each line was earned by an actual debugging session:

- **gh-commit**: atomic conventional commits, never auto-push, and the commit-type reservation that pairs with #432's changelog filter: bare `ci:` type belongs to the pipeline and is filtered from changelog sections; real CI work ships as `fix(ci)`/`feat(ci)`/`docs(ci)`
- **docker-test**: the compose smoke flow, the offline self-containment check (catches Bun's runtime auto-install quietly fetching packages on cold start), the dummy-env recipe, and the recorded traps: `docker run --env-file` keeps inline comments (NaN values), secret contents don't bust the build cache, dummy-built web images must never be promoted
- **ignore-sync**: `.dockerignore` doesn't inherit `.gitignore`, and unmirrored entries have cost gigabytes of build context downstream (a 23GB `.turbo` once entered a context); the skill keeps them line-for-line mirrored, with the env inversion now classified as drift rather than policy (it existed to feed the `COPY .env*` that #434 removes)

All three are generic: no downstream-specific content survives a grep.

**Docs**: `resources/ai-skills.mdx` updated with docker-test and ignore-sync sections, the Git Commit entry repointed to the real flat path (the doc referenced `.agents/skills/git/commit/`, which does not exist on canary), and the skill-structure layout corrected to flat. Note: #433 touches the same file's build-graph section; whichever merges second resolves a trivial conflict.

## Verification

- Content grep: zero downstream references
- Each skill's claims trace to verified behavior in the downstream history (offline check caught a real auto-install incident; the ignore-sync audit diff is the documented two-label expected output)

### DIFF
--- a/.agents/skills/docker-test/SKILL.md
+++ b/.agents/skills/docker-test/SKILL.md
@@ -1,0 +1,63 @@
+---
+name: docker-test
+description: Build and smoke-test the Docker images with docker compose. Use when touching Dockerfiles, the bundle build, compose config, or anything that changes how containers are packaged.
+---
+
+# Docker Testing
+
+Canonical flow (full stack, mirrors the production images):
+
+```bash
+docker compose build --no-cache && docker compose up
+```
+
+Scope to one service while iterating:
+
+```bash
+docker compose build --no-cache api && docker compose up -d api
+curl -sf --retry 30 --retry-delay 1 --retry-connrefused http://localhost:4000/api/health
+docker compose logs -f api
+docker compose down
+```
+
+## .env requirements
+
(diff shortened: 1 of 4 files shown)


nrjdalal:
Closing: new skills are coming soon, so this is no longer required.

PR #436, opened 2026-06-06
feat(web): self-host fonts via next/font/local with mandatory weight ranges

nrjdalal:
## Summary

The four brand fonts move from `@fontsource-variable/*` CSS imports to vendored woff2 + `next/font/local`, gaining build-time preload links, hashed immutable URLs, and size-adjusted fallback faces (the CLS mechanism, which is compiler-generated and architecturally unavailable to fontsource). Four dependencies removed; italics vendored for the prose faces (DM Sans, Newsreader).

**The trap this PR is fenced against**: `localFont` without an explicit `weight` defaults the face to 400, so every bold glyph silently renders as faux-bold synthesis. No build error, no CLS movement, nothing catches it except comparing against what the deleted fontsource CSS used to declare. All four faces declare their full variable ranges, and the new `fonts` skill marks the range mandatory with the exact grep to fetch it from the fontsource CDN.

**This supersedes the #383 approach.** Re-tested 2026-06-06 on Next 16.2.6 by replicating that PR's diff verbatim: prod builds are fixed upstream (preload hashes match CSS references byte for byte), but in dev the `*.woff2` asset rule still rewrites fontsource's CSS `url()`s to unservable node_modules paths, so every font 404s and dev runs on system fonts. A dev-broken approach can't ship regardless of prod; `next/font/local` works in both and adds the fallback metrics #383 never could. Detailed retest notes are on #383.

## Changes

- `web/next/src/fonts/`: six latin variable woff2 vendored (~315 KB; CDN-fetched, no packages)
- `web/next/src/lib/fonts.ts`: four `localFont` definitions with weight ranges; Newsreader uses Times New Roman fallback metrics
- `layout.tsx`: variables + `antialiased` consolidated on `<html>` via `cn`
- `globals.css`: fontsource imports removed, the four font roles chain to the generated variables
- Dependencies: four `@fontsource-variable/*` packages removed from web/next and the catalog
- New `fonts` skill: the swap recipe (CDN fetch, mandatory weight ranges, role wiring, removal steps); theming docs updated

## Verification

- Production build in this clone: all six `@font-face` ranges present in emitted CSS (`100 1000` ×2, `200 800` ×2, `100 800`, `400 700`), six `<link rel="preload" as="font">` in prerendered HTML with hashed URLs
- Downstream (dalonic/cafe) runs this pattern in production: preloads observed on deployed HTML, fallback faces observed in emitted CSS, CLS 0 across pages


nrjdalal:

PR #437, opened 2026-06-06
perf(api): cache auth sessions in a signed cookie

nrjdalal:
## Summary

Authenticated requests stop hitting Postgres on every call: better-auth's `cookieCache` stores session data in a short-lived signed cookie (300s window, explicit because that number is the revocation window and deserves to be a visible decision).

**The half that's easy to miss**: enabling the config alone is placebo. The auth middleware's server-side `auth.api.getSession({ headers })` silently discards better-auth's `Set-Cookie` response, so the cache cookie never reaches the client and every request still hits the database while looking perfectly healthy. The middleware switches to `returnHeaders: true` and forwards each cookie via `c.header("Set-Cookie", cookie, { append: true })`, which also forwards cookie-clearing headers on expired sessions for free.

**The tradeoff, stated**: revoked sessions stay valid on the cached cookie for up to the window (admin revocations, password-change-revokes-others, and bans all inherit it). `disableCookieCache: true` per-call is the documented escape hatch for revocation-sensitive operations; a Redis `secondaryStorage` is the multi-instance path if that ever matters.

## Changes

- `packages/auth`: `session.cookieCache` enabled, `maxAge: 300`
- `api/hono` auth middleware: `getSession` with `returnHeaders: true`, Set-Cookie forwarded
- `manage/authentication.mdx`: new Cookie Cache section covering the forwarding requirement and the revocation-window tradeoff

## Verification

- Typechecked in this clone against better-auth ^1.6.11 (the `returnHeaders` overload)
- Downstream (dalonic/cafe) verified adversarially: sign in, prime the cache cookie, revoke the session (row deleted), replay the cached cookie → 200 with the row gone, proving Postgres is bypassed; browser e2e confirmed the cookie pair primes through the dashboard and clears fully on logout
- Downstream also observed the placebo failure mode directly: with config-only, the `session_data` cookie never appears

### DIFF
--- a/api/hono/src/middlewares/auth.ts
+++ b/api/hono/src/middlewares/auth.ts
@@ -13,7 +13,16 @@ const userRateLimiter = createRateLimiter({
 })
 
 export const authMiddleware = createMiddleware<{ Variables: Session }>(async (c, next) => {
-  const session = await auth.api.getSession({ headers: c.req.raw.headers })
+  const { headers, response: session } = await auth.api.getSession({
+    headers: c.req.raw.headers,
+    returnHeaders: true,
+  })
+
+  // forward the session_data cookie so the cookie cache actually primes;
+  // getSession without this silently drops better-auth's Set-Cookie
+  for (const cookie of headers.getSetCookie()) {
+    c.header("Set-Cookie", cookie, { append: true })
+  }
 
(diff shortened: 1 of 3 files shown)


nrjdalal:

PR #451, opened 2026-06-11
feat(web): expand per-workspace MCP post (GH_TOKEN fix, hardening, prior art)

nrjdalal:
Revision of the per-workspace MCP/identity post, incorporating deep-research feedback.

## Correctness
- **`GH_TOKEN` precedence fix** in the Layer 3 hook: export/unset both `GH_TOKEN` and `GITHUB_TOKEN` so an inherited `GH_TOKEN` can't make `gh` disagree with the MCP header. (Same fix applied to the live `mcp-tokens.zsh`.)
- Tightened the env-var gotcha: distinguishes **unset** (config parse fails) from **set-but-empty** (`Bearer ` → HTTP 400), instead of "always HTTP 400".

## Trust / positioning
- "Project vs. workspace" note: clarifies this is an *ancestor* `.mcp.json`, and that the cross-repo walk-up is observed behavior (verified on Claude Code 2.1.173 via `claude mcp list`), not documented.
- Layer 2 caveat: unique names isolate *credentials*, not *access*; orgs should match URL/command.
- "Prior Art, and What's Different" + "When You Don't Need This" sections.
- Expanded the token-exposure gotcha into a hardening note (prompt-injection framing, read-only default, write wrapper, separate OS user).

## Additions
- `headersHelper` MCP-only variant (token never enters the shell) with the tradeoff vs the env-var approach.
- "The Layers, At a Glance" token/source matrix.
- Copy-paste verify script (incl. `GH_TOKEN`/`GITHUB_TOKEN`).
- Sharpened description toward race-free-across-concurrent-sessions (title unchanged).

## Summary by CodeRabbit

* **Documentation**
  * Rewrote intro and added explicit assumptions/examples for macOS zsh workflows
  * Clarified layered identity model, isolation limits, and GitHub auth resolution (paired tokens, HTTPS vs SSH)
  * Updated token handling guidance (paired unset/export behavior) and added an MCP-only header helper alternative
  * Added macOS keychain note, a one-command verification snippet, a summary identity table, and expanded troubleshooting/gotchas

### DIFF
--- a/web/next/content/blog/mcp-per-workspace.mdx
+++ b/web/next/content/blog/mcp-per-workspace.mdx
@@ -1,6 +1,6 @@
 ---
 title: "One Machine, Many Identities: Per-Workspace MCP and Git in Claude Code"
-description: "Route MCP servers, GitHub accounts, and git commit identity by directory, so work, client, and personal projects never bleed into each other. No manual switching, safe across concurrent terminals."
+description: "Make MCP servers, GitHub auth, and git commit identity follow the directory you're in, so concurrent Claude Code sessions never bleed work, client, and personal accounts into each other. No manual switching, no shared global state to race."
 ---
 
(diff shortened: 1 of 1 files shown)


nrjdalal:
Scrapping this expansion. Rewriting the post from scratch around the simpler unified single-hook design; fresh PR to follow.

PR #489, opened 2026-06-20
feat(auth): console roles via the better-auth admin plugin

nrjdalal:
Reinvents console access around the [better-auth Admin plugin](https://www.better-auth.com/docs/plugins/admin), replacing the custom `console` user field.

## Model

- `/console` is gated on **`user.role === "admin"`** (`web/next/src/lib/auth/console.ts`); every other account is the plugin's default `user` role → `notFound()`.
- **Two roles for now** (`admin` + default `user`). Built on the plugin's access-control, so adding tiers later is just extending the role map, no rework.
- **No env var.** `user.role` is the single source of truth: there is no env-based access list and no sign-up promotion hook. Admins are set explicitly.
- **Immediate effect.** The gate reads the session with `disableCookieCache`, so a grant/revoke takes effect on the **next** `/console` request rather than waiting out the 300s `session.cookieCache`.

## Bootstrap & management

- **`console:roles` helper** (`.github/scripts/console-roles.ts`, run via `bun run console:roles`): `grant <email>` / `revoke <email>` / `list`. Writes `user.role` directly through `POSTGRES_URL` with parameterized, case-insensitive (`lower(email)`) queries.
- **Or** `bun run db:studio` / `UPDATE "user" SET role = 'admin' WHERE email = '…';`.
- **Or**, once an admin exists, the plugin's `setRole`, the basis for the Phase 2 console user-management dashboard.

## Changes

- **`packages/auth`**: add the `admin()` plugin; remove the `console` `additionalField`. No sign-up hook, no new dependencies.
- **`packages/db`**: add the plugin fields, `role` (default `user`), `banned`, `ban_reason`, `ban_expires` on `user` and `impersonated_by` on `session`; drop the legacy `console` column. Migrations **0003** (add) + **0004** (drop), generated in two non-ambiguous steps to avoid drizzle's rename prompt. **0004 backfills `role = 'admin'` from any non-empty `console` before dropping the column**, so existing admins on already-deployed instances (forks / live DBs) carry over.
- **agent sign-in** (`api/hono/src/routers/agents.ts`): grants `role: "admin"` to the local `AgentZero` user.
- **web get-session** (`web/next/src/lib/auth/index.ts`): forwards an optional `disableCookieCache` query to the Hono get-session route for the console gate.
- **Docs**: `manage/authentication.mdx`, `manage/database.mdx` (schema + migration table), `console/docs/index.mdx`, all updated to the role / manual-bootstrap model.

## Verification

- `bun run check-types` 9/9; `oxlint` + `oxfmt --check` clean; full `bun run build` passes (incl. the pre-commit hook).
- Migrations applied to the dev DB; stack restarted; **browser-verified**: signing in as the local agent (now `role: "admin"`) renders `/console`. Logged-out / non-admin accounts hit the `null → notFound()` deny branch.
- `console:roles list` / `grant` exercised against the dev DB.

## Notes

- **Per-request DB read on the console gate** is deliberate (`disableCookieCache`), the tradeoff for immediate grant/revoke. The `(protected)` layout still uses the cached path. To cut an account off everywhere immediately, ban it or revoke its sessions.
- The Admin plugin also mounts `banUser`, `impersonate`, and `listUsers` (admin-gated), expected, and the basis for the Phase 2 dashboard.

## Summary by CodeRabbit

* **New Features**
  * Implemented role-based admin access system, replacing the previous console flag approach.
  * Added command-line tool for managing admin privileges (grant, revoke, and list operations).
  * Admin role changes now take effect immediately on the next request with no delay.

* **Documentation**
  * Updated guides on console access requirements and admin role management workflows.
  * Documented bootstrap process for initial admin setup and ongoing role administration.

### DIFF
--- a/packages/auth/src/index.ts
+++ b/packages/auth/src/index.ts
@@ -13,7 +13,12 @@ import {
 import { env } from "@packages/env/auth"
 import { betterAuth } from "better-auth"
 import { drizzleAdapter } from "better-auth/adapters/drizzle"
-import { openAPI as openAPIPlugin, organization as organizationPlugin } from "better-auth/plugins"
+import {
+  admin as adminPlugin,
+  openAPI as openAPIPlugin,
+  organization as organizationPlugin,
+} from "better-auth/plugins"
+import { eq } from "drizzle-orm"
 
 import { getCookieDomain, getCookiePrefix } from "@/lib/utils"
 
@@ -40,13 +45,18 @@ export const auth = betterAuth({
   onAPIError: {
     throw: true,
   },
-  user: {
-    additionalFields: {
-      // Null for normal users; a non-null value (e.g. "admin") grants access to the privileged /console area. `input: false` blocks self-assignment.
-      console: {
-        type: "string",
-        required: false,
-        input: false,
+  databaseHooks: {
+    user: {
+      create: {
+        // Promote the configured root admin to the `admin` role on first sign-up, so a fresh deploy gets console access without a manual DB edit.
+        after: async (createdUser) => {
+          if (
+            createdUser.emailVerified &&
+            env.CONSOLE_ADMIN_EMAILS.includes(createdUser.email.toLowerCase())
+          ) {
+            await db.update(user).set({ role: "admin" }).where(eq(user.id, createdUser.id))
+          }
+        },
       },
     },
   },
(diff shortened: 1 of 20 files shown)


packages/auth/src/index.ts
```diff
@@ -61,6 +55,7 @@ export const auth = betterAuth({
     organizationPlugin({
       teams: { enabled: true },
     }),
+    adminPlugin(),
```
nrjdalal, line 58: add this in alphabatical order

PR #530, opened 2026-06-21
feat(auth): make GitHub and Google OAuth optional; UI shows only configured providers

nrjdalal:
## What

Make each social OAuth provider independently **optional**, and have the sign-in UI render only the providers that are actually configured. A fresh fork can ship with GitHub, Google, both, or neither (magic link always works) without editing code or hitting env-validation errors.

## Why

An upstream audit of the dalonic forks surfaced this as the one genuinely upstreamable delta: the starter couldn't boot unless both GitHub and Google creds were present, and the login dialog hardcoded both buttons regardless of what was configured.

## How

Single source of truth = the API's env.

- `packages/env/src/auth.ts`: `GITHUB_*` / `GOOGLE_*` are now `.optional()`.
- `packages/auth/src/index.ts`: `socialProviders` is built conditionally (a provider registers only when both of its creds are set), and a derived `enabledSocialProviders` list is exported.
- `api/hono/src/routers/auth.ts`: public `GET /api/auth/providers` (after `get-session`, before the better-auth `/*` catch-all) returns that list, RPC-typed like `apiClient.waitlist`.
- `web/next/src/components/access.tsx`: fetches the list with react-query (`staleTime: Infinity`, prefetched on mount) and renders only the available buttons. The dialog content mounts on open and the list is prefetched, so the buttons paint with the final set on first open (no flash/CLS). A failed fetch throws so react-query retries instead of silently degrading to magic-link-only; the "OR" divider hides in production builds when nothing alternative is available. Magic-link email stays always-on.
- Docs + `.env.example` updated (`manage/authentication`, `manage/environment`), including steps for adding a brand-new provider (extend the `SocialProvider` union + add a gated button in `access.tsx`).

Toggling either built-in provider is just an env change; a brand-new provider also needs the small code change documented above.

## Verification

- `oxfmt`, `oxlint`, and all 12 `check-types` tasks pass (incl. `@web/next` `tsc --noEmit`, confirming the `apiClient.auth.providers.$get()` RPC type resolves).
- Runtime "hide" path exercised end to end (backed up `.env`, toggled creds, restored byte-for-byte with md5 match):

| `.env` state | `GET /api/auth/providers` | Login dialog |
| --- | --- | --- |
| GitHub + Google | `["github","google"]` | both buttons |
| Google blanked | `["github"]` | GitHub only |
| Both blanked | `[]` | no social buttons (email + dev agents-login remain) |
| Restored | `["github","google"]` | both buttons |

- Browser-confirmed in a real Chromium session (agent-browser) for the github-only and none-configured states.

## Summary by CodeRabbit

* **Improvements**
  * OAuth provider credentials for GitHub and Google are now optional; authentication sign-in buttons display only when providers are configured.

* **Documentation**
  * Updated authentication and environment setup guides to clarify optional OAuth configuration and conditional provider button behavior.

### DIFF
--- a/api/hono/src/routers/auth.ts
+++ b/api/hono/src/routers/auth.ts
@@ -1,6 +1,7 @@
-import { auth } from "@packages/auth"
+import { auth, enabledSocialProviders } from "@packages/auth"
 import { Hono } from "hono"
 
 export const authRouter = new Hono()
+  .get("/providers", (c) => c.json({ data: { providers: enabledSocialProviders } }))
   .get("/get-session", (c) => auth.handler(c.req.raw))
   .on(["GET", "POST"], "/*", (c) => auth.handler(c.req.raw))
--- a/api/hono/src/routers/auth.ts
+++ b/api/hono/src/routers/auth.ts
@@ -2,6 +2,6 @@ import { auth, enabledSocialProviders } from "@packages/auth"
 import { Hono } from "hono"
 
 export const authRouter = new Hono()
-  .get("/providers", (c) => c.json({ data: { providers: enabledSocialProviders } }))
   .get("/get-session", (c) => auth.handler(c.req.raw))
+  .get("/providers", (c) => c.json({ data: { providers: enabledSocialProviders } }))
   .on(["GET", "POST"], "/*", (c) => auth.handler(c.req.raw))
--- a/web/next/src/components/access.tsx
+++ b/web/next/src/components/access.tsx
@@ -3,6 +3,7 @@
 import { site } from "@packages/config/site"
 import { RiGithubFill, RiGoogleFill, RiLayoutGridFill, RiLoaderLine } from "@remixicon/react"
 import { useForm } from "@tanstack/react-form"
+import { useQuery } from "@tanstack/react-query"
 import { usePathname } from "next/navigation"
 import { useEffect, useState } from "react"
 import { toast } from "sonner"
@@ -18,6 +19,7 @@ import {
 } from "@/components/ui/dialog"
 import { Field, FieldError, FieldGroup, FieldLabel } from "@/components/ui/field"
 import { Input } from "@/components/ui/input"
+import { apiClient } from "@/lib/api/client"
 import { authClient } from "@/lib/auth/client"
 import { config } from "@/lib/config"
 
@@ -32,6 +34,20 @@ export function Access() {
   // Next inlines NODE_ENV at build time: "development" only under `next dev`,
   // "production" for any `next build`. Auto-hides in deployments.
   const isDev = process.env.NODE_ENV === "development"
+
+  // Render only the social buttons for providers the API actually has configured (GET /api/auth/providers).
+  const { data: providers } = useQuery({
+    queryKey: ["auth-providers"],
+    queryFn: async () => {
+      const res = await apiClient.auth.providers.$get()
+      if (!res.ok) return [] as ("github" | "google")[]
+      const { data } = await res.json()
+      return data.providers
+    },
+  })
+  const githubEnabled = providers?.includes("github") ?? false
+  const googleEnabled = providers?.includes("google") ?? false
+  const hasAlternatives = isDev || githubEnabled || googleEnabled
 
   useEffect(() => {
     setLoader(null)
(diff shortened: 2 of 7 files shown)


nrjdalal:
## Runtime verification (the "hide" path)

Backed up `.env`, toggled the OAuth creds, and verified end to end at the API, RPC, and browser levels, then restored `.env` byte-for-byte (md5 match).

| `.env` state | `GET /api/auth/providers` | Login dialog |
| --- | --- | --- |
| GitHub + Google set | `{"data":{"providers":["github","google"]}}` | both buttons |
| Google blanked | `{"data":{"providers":["github"]}}` | GitHub only, no Google |
| Both blanked | `{"data":{"providers":[]}}` | no social buttons (email + dev agents-login remain) |
| Restored | `{"data":{"providers":["github","google"]}}` | both buttons |

Browser-confirmed in a real Chromium session via agent-browser (screenshots captured for the github-only and none-configured states). The provider buttons appear/disappear purely from the API env, no web-side config. Note: the "OR" divider only hides in production builds, since the dev-only agents-login button keeps an alternative present under `next dev`.

api/hono/src/routers/auth.ts
```diff
@@ -1,6 +1,7 @@
-import { auth } from "@packages/auth"
+import { auth, enabledSocialProviders } from "@packages/auth"
 import { Hono } from "hono"
 
 export const authRouter = new Hono()
+  .get("/providers", (c) => c.json({ data: { providers: enabledSocialProviders } }))
   .get("/get-session", (c) => auth.handler(c.req.raw))
```
nrjdalal, line 5: change the order, get-session first then providers?

web/next/src/components/access.tsx
```diff
@@ -33,6 +35,20 @@ export function Access() {
   // "production" for any `next build`. Auto-hides in deployments.
   const isDev = process.env.NODE_ENV === "development"
 
+  // Render only the social buttons for providers the API actually has configured (GET /api/auth/providers).
+  const { data: providers } = useQuery({
```
nrjdalal, line 39: won't this call make things slower?

nrjdalal:
Thanks for the thorough review. Addressed in `d7b8938`.

**#1 (flash / CLS) — fixed without a skeleton.** Root cause re-examined: the buttons live inside `DialogContent`, and the dialog (`@base-ui/react/dialog`, no `keepMounted`) mounts its content only when it **opens**, while the providers query fires on component **mount** (page load). So by the time the user clicks Login, the list is already resolved and the dialog paints the final buttons on first open. No flash on the configured path, and no layout to reserve. I first tried a loading skeleton, but as noted that just relocates the shift (skeleton then collapses to fewer/zero buttons), so it was removed. Added `staleTime: Infinity` since the list is deploy-static (also kills any refetch flicker).

**#2 (API-blip fallback) — improved.** The `!res.ok` path now `throw`s instead of returning `[]`, so react-query retries the transient blip (default backoff) rather than silently sticking at magic-link-only. On a hard failure the buttons stay hidden (graceful), which is correct since social sign-in would fail against an unreachable API anyway. Chose retry + graceful-hide over an explicit error affordance on the login surface.

**#3 (docs) — fixed.** Softened "the UI follows automatically" to scope it to toggling the two built-in providers, and the "Adding a provider" steps now include extending the `SocialProvider` union and adding the gated button in `access.tsx` (with a note that skipping it leaves a server-enabled provider with no button).

**Minor:**
- `staleTime: Infinity` added (your nit).
- `enabled: open` deliberately **not** applied: scoping the query to dialog-open would reintroduce the flash #1 is about, since the fetch would then start on open instead of prefetching on mount. The tradeoff is one tiny cached GET per session on home load, which I think is worth it.
- Button-array refactor left as-is to keep the diff focused; noted as reasonable future polish.

The runtime matrix (incl. the github-only / none browser screenshots) from the earlier comment still holds; this change doesn't alter the resolved-state rendering.

api/hono/src/routers/auth.ts
nrjdalal: Done in 8247cb6. It's functionally identical either way: `/get-session` and `/providers` are distinct exact paths, so their relative order doesn't affect matching. The only ordering that matters is the better-auth `/*` catch-all staying last, which it does. Reordered to keep get-session first.

web/next/src/components/access.tsx
nrjdalal: No meaningful slowdown. It's a single tiny GET (a short array), fired once on mount with `staleTime: Infinity`, so the QueryClient caches it for the whole session: no refetch across client-side navigations, only a hard reload refetches. It's non-blocking, the page paints and is interactive immediately while the query resolves in the background, and by the time someone clicks Login the dialog renders from cache. The only cost is one small cached GET on the first public-page load.

If you'd rather not pay even that, `enabled: open` would defer the fetch to dialog-open, but that reintroduces the open-time flash we just removed, so I kept the prefetch. A build-time/SSR list isn't a clean fit since web and the API deploy separately (which is also why api-status and the waitlist count are client fetches).

PR #532, opened 2026-06-21
feat(auth): gate the magic-link email field on the plugin being enabled

nrjdalal:
## What

Follow-up to #530. The email/magic-link field in the sign-in dialog rendered unconditionally, even though the server-side `magicLink` plugin is **not** registered by default. Out of the box it was a dead control: submitting an email just errored. This gates it the same way social providers are gated, so the dialog only shows providers that actually work.

## How

Single source of truth = which providers/plugins are configured on the API. `GET /api/auth/providers` returns one unified list, e.g. `["github","google","magic-link"]`.

- `packages/auth/src/index.ts`: `magicLinkEnabled` is derived from the constructed instance (`auth.options.plugins`), so registering the `magicLink` plugin flips it automatically (no separate flag to keep in sync). Plugins stay inline in `betterAuth()` to preserve the `$Infer.Session` augmentations. `enabledProviders` = `enabledSocialProviders` + `magic-link` (when enabled).
- `api/hono/src/routers/auth.ts`: the route returns `{ providers: enabledProviders }` — one array typed `AuthProvider = "github" | "google" | "magic-link"`, rather than a providers array plus a separate `magicLink` boolean.
- `web/next/src/components/access.tsx`: reads the one list via `includes()`; the email form renders only when `magic-link` is present, the "OR" divider shows only when both an email method and social/agents methods exist (no orphan divider), and a small "No sign-in options are configured yet." note shows when nothing is enabled.
- Docs (`manage/authentication`) updated.

## Verification

- `oxfmt`, `oxlint`, and all 12 `check-types` tasks pass (incl. `@web/next` — Session inference and the `AuthProvider[]` RPC type both resolve).
- Runtime: `auth.options.plugins` → `["open-api","organization","admin"]` (no `magic-link`); `GET /api/auth/providers` → `{"data":{"providers":["github","google"]}}`.
- Browser (agent-browser, `next dev`): with magic link off, the dialog shows social buttons + dev agents-login and **no email field and no "OR" divider** (screenshot captured). The reverse ("email appears once the plugin is registered") follows from the runtime-confirmed derive plus the trivial `includes("magic-link")` gate; not exercised live since wiring the plugin (a `sendMagicLink` sender) is a separate task.

## Summary by CodeRabbit

* **New Features**
  * Added magic-link sign-in option, conditionally available when configured on the server.

* **Documentation**
  * Updated authentication configuration guide to clarify magic-link setup and provider detection.

* **Refactor**
  * Improved dynamic detection and rendering of available authentication providers in the login interface.

### DIFF
--- a/api/hono/src/routers/auth.ts
+++ b/api/hono/src/routers/auth.ts
@@ -1,7 +1,9 @@
-import { auth, enabledSocialProviders } from "@packages/auth"
+import { auth, enabledSocialProviders, magicLinkEnabled } from "@packages/auth"
 import { Hono } from "hono"
 
 export const authRouter = new Hono()
   .get("/get-session", (c) => auth.handler(c.req.raw))
-  .get("/providers", (c) => c.json({ data: { providers: enabledSocialProviders } }))
+  .get("/providers", (c) =>
+    c.json({ data: { providers: enabledSocialProviders, magicLink: magicLinkEnabled } }),
+  )
   .on(["GET", "POST"], "/*", (c) => auth.handler(c.req.raw))
--- a/api/hono/src/routers/auth.ts
+++ b/api/hono/src/routers/auth.ts
@@ -1,9 +1,7 @@
-import { auth, enabledSocialProviders, magicLinkEnabled } from "@packages/auth"
+import { auth, enabledProviders } from "@packages/auth"
 import { Hono } from "hono"
 
 export const authRouter = new Hono()
   .get("/get-session", (c) => auth.handler(c.req.raw))
-  .get("/providers", (c) =>
-    c.json({ data: { providers: enabledSocialProviders, magicLink: magicLinkEnabled } }),
-  )
+  .get("/providers", (c) => c.json({ data: { providers: enabledProviders } }))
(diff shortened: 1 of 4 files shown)


nrjdalal:
Thanks for the thorough review and for externally confirming the `"magic-link"` plugin id. Both optional items addressed in 07de479.

- **`hasSocial` → `hasAlternatives`**: agreed, it included `isDev` (the dev agents-login), so it read narrower than it was. Renamed.
- **All-off production empty state**: documented in `manage/authentication.mdx` as an expected default — a deployed fork with no OAuth and no magic link shows the "No sign-in options are configured yet." note (dev agent sign-in is hidden in prod), so a fork must configure at least one method before shipping login.

On the test-coverage suggestion (a story/assertion rendering `access.tsx` across the four provider states): agreed it's the kind of gating logic that can silently drift. The repo has no web/next test harness yet (CAF-95 is a deliberate deferral), so I'll leave it for when that lands rather than add a one-off here. The two prod-only states remain validated by inspection + types + the now-confirmed plugin id.

check-types 12/12, lint + format clean.

PR #533, opened 2026-06-21
refactor(web): standardize TanStack Form practices across forms

nrjdalal:
## Why

Evaluated our TanStack Form usage against the shadcn [TanStack Form guide](https://ui.shadcn.com/docs/forms/tanstack-form.md). Configuration and component composition already match the guide 1:1. This PR closes the few **internal-consistency** gaps so all three forms (`access`, `waitlist`, `org-switcher`) follow one house standard.

## House standard

- `useForm({ defaultValues, validators, onSubmit })` with Zod (v4) schemas
- `form.Field` with the JSX-children render-prop
- `isInvalid = field.state.meta.isTouched && !field.state.meta.isValid`
- `data-invalid` on `<Field>`, `aria-invalid` on the control
- `{isInvalid && <FieldError errors={field.state.meta.errors} />}`
- Submit via `onSubmit={(e) => { e.preventDefault(); form.handleSubmit() }}`
- **Submit-disabled UI driven by `form.state.isSubmitting`** (exception: surfaces with multiple concurrent async actions, like `access`'s email + GitHub + Google, may keep explicit state)
- `autoComplete` set on every input

## Changes

| File | Change | Reason |
| --- | --- | --- |
| `org-switcher.tsx` | add `onBlur` validator | match the `{ onSubmit, onChange, onBlur }` set `access` + `waitlist` use |
| `waitlist/page.tsx` | replace manual `useState(loading)` + `try/finally` with `form.state.isSubmitting` | align on the TanStack-native submit state `org-switcher` already ships; removes a flag that can drift |
| `access.tsx`, `waitlist`, `org-switcher` | add `autoComplete` (`email` on email fields, `off` on org name) | the guide sets `autoComplete` on every input; UX + a11y |

## Deliberately not changed

- **Render-prop syntax** stays JSX-children (cleaner than the guide's `children={}` attribute, which needs `eslint-disable react/no-children-prop`); already consistent across all three forms
- **Zod v4** kept (guide examples are v3; it notes any Standard Schema lib works)
- **className strings** are per-form layout (Dialog / hero / sidebar), not a parity target
- **`field.tsx` internals** out of scope (the guide does not show the component source)

## Verification

- `oxlint` clean on all three files
- The `form.state.isSubmitting` swap is behavior-equivalent and mirrors the pattern `org-switcher` already ships on canary
[...]

PR #535, opened 2026-06-21
feat: { data, error } API response shape (unwrap client + OpenAPI error docs)

nrjdalal:
## Why

Make API responses a `{ data, error }` shape that's clean to consume on the client and fully documented in the API reference. (Combines the former #536.)

## What

**Client: `unwrap`** (`web/next/src/lib/api/client.ts`): wrap any RPC call, get `{ data, error }` (exactly one non-null), never throws.

```ts
const { data, error } = await unwrap(apiClient.health.$get())
```

- Reads `throw` on the error arm so React Query reports `isError`; the waitlist submit is a `useMutation` that throws in `mutationFn` and toasts in `onError` (no toast in the success path).
- The generic is `<R extends RpcResponse>` + `Awaited<ReturnType<R["json"]>>`, so it handles Hono's per-status `ClientResponse` union (e.g. the waitlist `POST`'s `200` + validator `400`).
- Better-auth calls are untouched (a separate client that already returns `{ data, error }`).

**Server: OpenAPI error docs** (`api/hono/src/lib/error.ts`): each route advertises only the statuses it can actually return. `429/500` apply to every route via `index.ts` `defaultOptions` (`globalErrorResponses`); auth routes (`/api/v1/*`) add `401` (`authErrorResponses`); validated routes (waitlist `POST`) add `400` (`validationErrorResponses`). A route's own `responses` merge over the defaults (**`200` first, errors grouped after**), each with its **own** code/message example. `AppType` is unchanged.

**Server: error handling** (`api/hono/src/lib/error.ts`): `errorHandler` maps `ZodError → 400 VALIDATION_ERROR` (with `issues`) and now honors `HTTPException`, returning its status instead of masking every non-Zod error as a `500`. So a malformed JSON or form-data body is `400 BAD_REQUEST`, not `500`. Route validators (waitlist `POST`) include the `issues` array, so a wrong or empty body shows the real cause instead of a bare message.

**Docs**: `type-safe-api.mdx`, `api-conventions.mdx`, `README.md`, the homepage sample, and the OpenAPI `x-codeSamples` show `unwrap`.

## Verification

- `check-types` pass (`api/hono` + `web/next`); lint + format clean; pre-commit build passed.
- Live spec (`:4000`), per-route reachable statuses only, each with a distinct example per status:
  - `GET /api/health` → `200, 429, 500`
  - `GET /api/v1/{session,user}` → `200, 401, 429, 500`
  - `GET /api/waitlist` → `200, 429, 500`
  - `POST /api/waitlist` → `200, 400, 429, 500`
- Live error behavior on `POST /api/waitlist`: malformed JSON → `400 BAD_REQUEST` (was `500`); wrong content-type or missing email → `400` with an `issues` array; bad email → `400` with `issues`; valid → `200`.

## Note

Touches `web/next/src/app/waitlist/page.tsx`, which PR #533 (TanStack Form parity) also touches; small merge reconciliation on whichever lands second.

## Summary by CodeRabbit

* **New Features**
  * Added a typed `unwrap` helper that returns `{ data, error }` (no throwing) for safer, consistent API response handling across the app.
  * Improved API error handling and OpenAPI-documented error responses, including standardized validation and auth error outputs.

* **Documentation**
  * Updated getting-started, API conventions, and landing page examples to use `unwrap` and demonstrate `{ data, error }`.
  * Expanded guidance on error codes, validation behavior, and documenting route error responses.

* **UI Improvements**
  * Refreshed waitlist and auth-related screens to use improved client response handling and clearer error toasts.

### DIFF
--- a/web/next/src/components/api-status.tsx
+++ b/web/next/src/components/api-status.tsx
@@ -2,17 +2,15 @@
 
 import { useQuery } from "@tanstack/react-query"
 
-import { apiClient } from "@/lib/api/client"
+import { apiClient, unwrap } from "@/lib/api/client"
 
 export function ApiStatus() {
   const { isLoading, isError } = useQuery({
     queryKey: ["api-health"],
     queryFn: async () => {
-      const res = await apiClient.health.$get()
-      if (!res.ok) {
-        throw new Error("Systems are facing issues")
-      }
-      return res.json()
+      const { data, error } = await unwrap(apiClient.health.$get())
+      if (error) throw new Error("Systems are facing issues")
+      return data
     },
     refetchInterval: 30000,
   })
--- a/web/next/src/components/api-status.tsx
+++ b/web/next/src/components/api-status.tsx
@@ -9,7 +9,7 @@ export function ApiStatus() {
     queryKey: ["api-health"],
     queryFn: async () => {
       const { data, error } = await unwrap(apiClient.health.$get())
-      if (error) throw new Error("Systems are facing issues")
+      if (error) throw new Error(error.message)
       return data
     },
     refetchInterval: 30000,
--- a/README.md
+++ b/README.md
@@ -67,11 +67,10 @@ ZeroStarter uses [Hono RPC](https://hono.dev/docs/guides/rpc) for end-to-end typ
 - **API Docs**: an interactive reference is served at `/api/docs` (OpenAPI spec at `/api/openapi.json`).
 
 ```ts
-import { apiClient } from "@/lib/api/client"
+import { apiClient, unwrap } from "@/lib/api/client"
 
-// Fully typed request and response
-const res = await apiClient.health.$get()
-const { data } = await res.json()
+// Fully typed { data, error } result (never throws)
+const { data, error } = await unwrap(apiClient.health.$get())
 ```
 
 ## 🔥 Why ZeroStarter?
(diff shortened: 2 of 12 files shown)


web/next/src/components/api-status.tsx
```diff
@@ -2,17 +2,15 @@
...
   const { isLoading, isError } = useQuery({
     queryKey: ["api-health"],
     queryFn: async () => {
-      const res = await apiClient.health.$get()
-      if (!res.ok) {
-        throw new Error("Systems are facing issues")
-      }
-      return res.json()
+      const { data, error } = await unwrap(apiClient.health.$get())
+      if (error) throw new Error("Systems are facing issues")
```
nrjdalal: are error supposed to be overriden like this?

nrjdalal: Good catch — it was discarding the typed error. Fixed in 9a37d85: both queryFns now `throw new Error(error.message)` (here and the providers query in `access.tsx`), so the real message from `unwrap` flows through instead of a hardcoded string.

The user-facing "Systems are facing issues" / "Failed to load auth providers" copy lives in the component JSX (the `isError` branch), which is what actually renders — so the thrown message is only for logs/devtools, and it is now the real one.

PR #536, opened 2026-06-21
feat(api): advertise the standard error responses in the OpenAPI docs

nrjdalal:
## Why

The OpenAPI / Scalar docs at `/api/docs` only listed each route's declared `200` — the standard `jsonError` error envelope (`{ error: { code, message } }`) was invisible. This advertises it on every route, with a correct per-status example.

Follow-up to #535 (the client `{ data, error }` / `unwrap` work).

## What

- `lib/error.ts`: add the `errorEnvelope` zod schema as the single source of the error shape (`jsonError` / `errorHandler` unchanged).
- `index.ts`: `openAPIRouteHandler` `defaultOptions: { GET/POST: { responses } }` adds `400/401/403/404/429/500` to every documented route, **each with its own code/message example** (e.g. `401 → UNAUTHORIZED / "Unauthorized"`, `400 → VALIDATION_ERROR / "Invalid request payload"`) — not one shared `INTERNAL_SERVER_ERROR` body.

`AppType` is unchanged — docs-only; the client (`unwrap`) returns a fixed `ApiError` and needs no typed error arm.

## Ordering

OpenAPI `responses` are integer-keyed, so they serialize in ascending status order — **`200` first, then the error codes grouped after** (`400, 401, 403, 404, 429, 500`).

## Verification

- `check-types` clean (`api/hono`); lint + format clean; pre-commit build passed.
- Generated spec confirms the order `200, 400, 401, 403, 404, 429, 500` and a distinct example per status.

## Note

Touches `api/hono/src/index.ts`, which #535 also edits (the health `x-codeSamples` string) — different regions, so it auto-merges; no real conflict.

## Summary by CodeRabbit

* **Documentation**
  * Standardized API error responses with consistent OpenAPI documentation for GET and POST endpoints, ensuring uniform error format across the API.

### DIFF
--- a/api/hono/src/index.ts
+++ b/api/hono/src/index.ts
@@ -8,7 +8,7 @@ import { cors } from "hono/cors"
 import { logger } from "hono/logger"
 import { z } from "zod"
 
-import { errorHandler, jsonError } from "@/lib/error"
+import { errorEnvelope, errorHandler, jsonError } from "@/lib/error"
 import { rateLimiterMiddleware } from "@/middlewares"
 import { agentsRouter, authRouter, v1Router, waitlistRouter } from "@/routers"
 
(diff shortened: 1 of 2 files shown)


nrjdalal:
Merged into #535 — the OpenAPI error responses now ship in the same PR, with errorEnvelope and the errorResponses builder living in api/hono/src/lib/error.ts (index.ts only wires defaultOptions).

PR #552, opened 2026-06-24
feat(api): centralize Hono route plumbing in defineRoute/ok/validate helpers

nrjdalal:
## What

A minimal way to write Hono routes here: OpenAPI docs + RPC types preserved, per-route footprint cut to intent, concerns separated. New module `api/hono/src/lib/route.ts` (39 lines):

- **`defineRoute({ tags, description, output, input?, auth? })`** — builds the route's OpenAPI doc. Wraps `output` in the `{ data }` envelope for the `200`, adds `400` when `input` is given and `401` when `auth: true`. `429`/`500` stay global (already applied by `openAPIRouteHandler`'s `defaultOptions`).
- **`ok(c, data)`** — typed success envelope. The handler still calls `c.json` (via `ok`), so Hono's **RPC type inference stays anchored** and the client keeps inferring `{ data: T }`.
- **`validate(schema)`** — body validator placed *directly* in the chain (not spread), so `c.req.valid("json")` stays typed; its failures **throw**, so `onError` shapes every validation `400` in one place.

Converted `waitlist`, `v1` (auth), `/health`, `/`, `/headers`.

## Design decisions (grilled)

- **Helper stops short of owning `c.json`.** A full handler-wrapping factory erodes Hono's RPC inference — verified: spreading a middleware tuple collapsed `c.req.valid("json")` to `never`. So `ok()` keeps `c.json` at the route; `validate()` is a direct arg.
- **Validation funnels through `onError`** (one envelope, one test surface); drops the per-route custom message and fixes the `issues: result.error` shape bug.
- **`429/500` stay global**, routes only declare their own `400`/`401`.

## Verified

- `api/hono` + `web/next` type-check clean → RPC inference and `c.req.valid` typing both intact.
- Generated the live OpenAPI spec from the app:
  - `POST /api/waitlist` → `200, 400, 429, 500`; `200` schema = `{ data: { message } }`
  - `GET /api/v1/user` → `200, 401, 429, 500`
  - `GET /api/waitlist` / `GET /api/health` → `200, 429, 500`

## Footprint

Routers shed **163 lines** of plumbing for **29 lines** of declarations behind a **39-line** module (+70/−160 net).

## Trade-off / follow-ups

- The hand-written `x-codeSamples` (the `hono/client` snippet) are dropped — they were ~12 lines of per-route boilerplate and can return as an optional one-line `client` hint if we want them in Scalar.
- `agentsRouter` / `authRouter` not converted (auth is Better Auth's handler; agents is local-only) — the helper is ready for them.

## Summary by CodeRabbit

* **New Features**
  * Standardized API responses across multiple endpoints using a consistent `{ data: ... }` JSON envelope.
  * Updated OpenAPI/route documentation wiring for v1 session/user, health, and waitlist endpoints.
* **Bug Fixes**
  * Made request validation and error handling more consistent for form submissions and protected routes.
* **Documentation**
  * Improved route metadata definitions to better match actual response shapes (including waitlist and auth-related endpoints).

### DIFF
--- a/api/hono/src/index.ts
+++ b/api/hono/src/index.ts
@@ -9,11 +9,13 @@ import { logger } from "hono/logger"
 import { z } from "zod"
 
 import { errorHandler, globalErrorResponses, jsonError } from "@/lib/error"
-import { defineRoute, ok } from "@/lib/route"
+import { routeGroup } from "@/lib/route"
 import { rateLimiterMiddleware } from "@/middlewares"
 import { agentsRouter, authRouter, v1Router, waitlistRouter } from "@/routers"
 
 const BUILD_VERSION = getBuildVersion()
+
+const route = routeGroup({ tags: ["System"] })
 
 const app = new Hono()
 
@@ -35,18 +37,17 @@ app.onError(errorHandler)
 app.notFound((c) => jsonError(c, 404, "NOT_FOUND", "Not Found"))
 
 const routes = app
-  .get("/", (c) => ok(c, { version: BUILD_VERSION, environment: env.NODE_ENV }))
+  .get("/", (c) => c.json({ data: { version: BUILD_VERSION, environment: env.NODE_ENV } }))
   .get("/headers", (c) => {
     if (env.NODE_ENV !== "local" && env.NODE_ENV !== "development") {
       return jsonError(c, 403, "FORBIDDEN", "Forbidden")
     }
-    return ok(c, c.req.header())
+    return c.json({ data: c.req.header() })
   })
   .basePath("/api")
   .get(
     "/health",
-    defineRoute({
-      tags: ["System"],
+    route({
       description: "Get the system health",
       output: z.object({
         environment: z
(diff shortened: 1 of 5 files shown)


nrjdalal:
Follow-up cleanup (`1cc04b7`): added `scopeRoute({ tags, auth })` to pre-bind a router's shared `tags`/`auth`, so routes declare only what varies. `auth`'s `/providers` now returns via `ok()`. Types + OpenAPI re-verified (v1 still 200/401/429/500, waitlist POST still 200/400/429/500).

I explored documenting the request body from `input` too, but `describeRoute`'s `requestBody.schema` won't accept hono-openapi's `resolver()` (only `responses` do), and a cast would be a hack — left it out as a separate doc-completeness item.

nrjdalal:
Closing — abandoning this. For a starter, the explicit hono-openapi routes are the right call. The `routeGroup`/`ok` abstraction added a bespoke concept and fought Hono's types at three points: spreading middleware breaks `c.req.valid` inference, `resolver()` can't go in `describeRoute`'s `requestBody`, and `@hono/standard-validator`'s `result.error` isn't a `z.ZodError` so routing validation through `onError` crashes. Even the trimmed 'validation via onError' bit was a regression — the original `return jsonError(..., { issues: result.error })` is correct (`result.error` already is the issues array). `canary` keeps the idiomatic routes; nothing was merged.

PR #553, opened 2026-06-24
refactor(api): contained error handling + typed ErrorCode union end-to-end

nrjdalal:
Single PR for the API error-handling rework and its end-to-end typing (#554 folded in as `80e6f69`). Contained strictly to error handling + how it's consumed — no route/RPC/OpenAPI abstraction (that was #552, closed).

### 1 — Contain error shaping (`ApiError` + `onError`)
- **`ApiError extends HTTPException`** (`lib/error.ts`) carries the envelope's domain `code` + optional `extra`; throw it anywhere.
- **`onError` is the one switchboard** that turns an error into `{ error: { code, message, … } }`: `ApiError` → `ZodError` → `HTTPException` (status-mapped) → `500`. Routes stop hand-building envelopes and **throw**: `/headers` (`HTTPException(403)`), `agents` `fail` (`ApiError(500, "AGENTS_LOGIN_FAILED")`), waitlist validator (`ApiError(400, "VALIDATION_ERROR", …, { issues })`).

### 2 — Type the `ErrorCode` union end-to-end
- A single, **alphabetized** `ErrorCode` union in `lib/error.ts` (used by `ApiError` / `jsonError` / `httpExceptionCodes`), re-exported from `@api/hono`.
- The web client's `error.code` (`lib/api/client.ts`) is now `ErrorCode | "NETWORK_ERROR" | "UNKNOWN_ERROR"` instead of `string` → autocomplete + exhaustive `switch`.

### 3 — Tidy frontend query consumption (folded from #554)
- `useQuery` + `unwrap` sites return the unwrapped `data` object rather than pre-extracting a field: `WaitlistCount` (`data` vs `data.count`) and `Access` (`data` vs `data.providers`); `ApiStatus` already did. No behavior change.

### Docs
- `web/next/content/docs/manage/api-conventions.mdx` synced to the throw/`onError` model (corrects the stale `jsonError`/`agents.ts` citation), documents `ApiError`, and alphabetizes the error-codes table to mirror the union (added the missing `ERROR` row).

## Grounding
- Hono's idiom is `throw` → central `onError` dispatching by `instanceof`; its custom-error example extends `HTTPException`. [HTTPException](https://hono.dev/docs/api/exception)
- The earlier "route validation through onError" crash is fixed correctly here by throwing our own `ApiError` (a real `Error`), not `@hono/standard-validator`'s non-`Error` `result.error`.

## Verified
[...]

PR #554, opened 2026-06-24
refactor(web): return unwrapped data from useQuery, not a pre-extracted field

nrjdalal:
## What

Make the `useQuery` + `unwrap` sites return the **unwrapped `data` object** and read the field in render, instead of pre-extracting a single field in the `queryFn`.

- `WaitlistCount` (`app/waitlist/page.tsx`): `queryFn` returns `data` (was `data.count`); render uses `data && data.count > 0` → `{data.count}+ …`.
- `Access` (`components/access.tsx`): `queryFn` returns `data` (was `data.providers`); consumers use `data?.providers.includes(…)`.
- `ApiStatus` already returned `data` — left as-is (now consistent across all sites).

## Why

The query holds the raw API payload and the component selects what it needs in render — no narrowing/massaging in the `queryFn`, type flows straight from `unwrap`.

## Notes
- No behavior change (same rendering conditions). `web` type-checks clean; pre-commit build green.
- Independent of #553 (touches only these two web components), so branched off `canary`.

### DIFF
--- a/web/next/src/app/waitlist/page.tsx
+++ b/web/next/src/app/waitlist/page.tsx
@@ -22,19 +22,19 @@ const formSchema = z.object({
 
 function WaitlistCount() {
   // the API returns a display-ready count (floored and rounded server-side)
-  const { data: count } = useQuery({
+  const { data } = useQuery({
     queryKey: ["waitlist-count"],
     queryFn: async () => {
       const { data, error } = await unwrap(apiClient.waitlist.$get())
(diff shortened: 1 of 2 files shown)


nrjdalal:
Folded into #553 to keep it a single PR (cherry-picked as `80e6f69` onto the error-handling branch). The query-return-data change is unrelated to the files in #553, so it merges cleanly alongside.

PR #579, opened 2026-06-28
refactor(web): cursor-pointer only on navigation, not action buttons

nrjdalal:
## cursor-pointer only on navigation

Applies the rule: `cursor-pointer` belongs only on path-changing navigation (links/anchors), never on action buttons. This matches the CSS spec / accessibility guidance (the pointer signals a link) and the Tailwind v4 + shadcn default (buttons use the arrow).

### Key finding
`buttonVariants` sets no cursor, and the browser already does the right thing: `<a href>` (including a Button rendered as `<Link>`) shows the pointer natively, `<button>` shows the arrow natively. So `cursor-pointer` was only ever *forcing* a pointer onto action buttons (against the rule) or sitting redundant on nav anchors. The pass removes every `cursor-pointer` class.

### Changes
- Removed `cursor-pointer` from all action controls: theme toggle, dialog/menu triggers, sidebar trigger, floating trigger, mobile-menu trigger, sign-in (magic-link / social / agents), sign-out, org-switch, create-org.
- Removed the now-redundant class from nav anchors (Dashboard-as-`Link`, the user-menu cross-link, the UserJot link) - they keep the pointer the `<a>` gives them for free.
- The docs search trigger (a readOnly `<input>`) uses `cursor-default` - a button-like control, so no pointer, but not a text field, so no I-beam.
- No `components/ui/*` files touched.

### Verified
- `rg cursor-pointer web/next/src` → zero matches.
- check-types 12/12, format clean.
- Browser (computed `cursor`): action buttons `default`, links `pointer`, search input `default`.

Standards: [CSS `cursor: pointer` is link-only](https://github.com/w3c/csswg-drafts/issues/7772), [Tailwind v4 dropped it from buttons](https://github.com/tailwindlabs/tailwindcss/pull/8962).

### DIFF
--- a/web/next/src/components/access.tsx
+++ b/web/next/src/components/access.tsx
@@ -85,9 +85,7 @@ export function Access() {
 
   return (
     <Dialog open={open} onOpenChange={setOpen}>
-      <DialogTrigger render={<Button className="w-24 cursor-pointer" variant="outline" />}>
-        Login
-      </DialogTrigger>
+      <DialogTrigger render={<Button className="w-24" variant="outline" />}>Login</DialogTrigger>
       <DialogContent className="max-w-md" initialFocus={false}>
         <DialogHeader className="sr-only">
           <DialogTitle className="text-center">Sign in/up</DialogTitle>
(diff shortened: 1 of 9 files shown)


nrjdalal:
Consolidated into #580 (cursor-pointer commit cherry-picked onto that branch).

PR #575, opened 2026-06-27
feat(web): DashboardShell + DashboardHeader page wrappers

nrjdalal:
## Dashboard page-layout wrappers

Central layout primitives so protected pages compose wrappers instead of hand-rolling spacing/sizing classes - reusable and non-drifting.

- **`DashboardShell`** (`components/dashboard/shell.tsx`): the page-content container; owns `mx-auto` + width + padding. `size` prop: `sm` (max-w-2xl), `md` (max-w-4xl), `lg` (default, max-w-6xl), `full`.
- **`DashboardHeader`** (`components/dashboard/header.tsx`): the `title` / `description` / `actions` row; owns the heading typography and spacing.

Usage:

```tsx
<DashboardShell size="sm">
  <DashboardHeader title="Passkeys" description="..." actions={<Button />} />
  {/* content */}
</DashboardShell>
```

### Also in this PR

- **Fills the shared `SidebarShell` `<main>`** (`flex min-w-0 flex-1 flex-col`). The bare `<main>` collapsed to content width, so `DashboardShell`'s `w-full` couldn't size or center; this is the prerequisite for the wrappers to work, and it benefits every protected/console page.
- **Wires the dashboard home** as the first consumer (was `return null`).
- **Docs**: added the `dashboard/` component group to project-structure.

### Verified

- `check-types` clean (12/12).
- Browser: dashboard home fills and centers (`mainW: 1024, shellW: 1024` at 1280px viewport), header renders.

### Notes

- Serves the console route group too (both use `SidebarShell`).
- The parked passkey PR (#574) adopts these wrappers when it resumes; its passkeys page currently hand-rolls the same layout.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added a new dashboard page layout for protected pages with a header, title, and welcome message.
  * Introduced a more flexible dashboard container with responsive sizing options.

* **Bug Fixes**
  * Improved sidebar/content layout so the main area sizes and scrolls more consistently.

* **Documentation**
  * Updated the project structure guide to include the new dashboard layout area.

PR #583, opened 2026-06-28
feat(web): route boundaries + standardize viewport idiom (svh top-level, flex-1 nested)

nrjdalal:
## FE foundations: route boundaries + viewport idiom

From the FE-refinements audit (`.github/audit/2026-06-29-fe-refinements.md`, included here).

### Route resilience
Error + loading boundaries built from `Empty` + `Spinner`, verified in-browser (the error renders *inside* the shell and fills the content pane):
- `(protected)` and `(console)` segment boundaries (page-level errors, in-shell).
- **Root `app/error.tsx`** - catches the route-group *layouts'* own `await getSession()` throw. A segment `error.tsx` can't catch its own layout (Next.js semantics), so the auth-failure path needs a parent boundary.
- **`app/global-error.tsx`** - the net for a root-layout throw (defines its own `<html>`/`<body>` + globals).
- Loading is scoped to `(protected)`/`(console)` (the data-heavy app surfaces); there is **no** root `app/loading.tsx` - a global full-viewport spinner is the wrong Suspense fallback for the prerendered marketing/content routes. Error coverage is global, loading is not; they needn't be symmetric.
- Shared `RouteError` / `RouteLoading` components; the segment files are thin wrappers so the boundaries can't drift.
- Per the Next.js 16.2 docs: error prop type is `Error & { digest?: string }`; kept the stable `reset` rather than the recommended-but-`unstable_`-prefixed `unstable_retry` (a forked starter shouldn't ship an unstable API).

### Viewport idiom standardized
- Top-level full-height surfaces (body, marketing pages, the `SidebarShell` root) use `min-h-svh` (was `dvh`; matches the shadcn sidebar).
- The shell `<main>` is `flex min-h-svh flex-1 flex-col`; nested surfaces fill it with `flex-1`. Recorded in the design skill.

### Verification

Boundaries verified in-browser (screenshots temporary, 72h):

| Throw origin | Boundary that catches it | Result |
|---|---|---|
| Page (segment) | segment `error.tsx` | renders **inside** the shell, fills the content pane |
| Route-group layout (`getSession`) | root `app/error.tsx` | **full-page** (no shell - the layout itself failed) |

Page error - in the shell (sidebar persists):

Segment error renders inside the shell, filling the content pane

Layout `getSession()` throw - caught by the root boundary, full-page:

Layout throw caught by root app/error.tsx, full-page with no shell

Next in the plan: dashboard/console content (gated on #575), then marketing polish.


web/next/src/app/global-error.tsx
```diff
@@ -0,0 +1,18 @@
...
+import "@/app/globals.css"
+
+export default function GlobalError(props: {
+  error: Error & { digest?: string }
+  reset: () => void
+}) {
+  return (
+    <html lang="en">
+      <body>
+        <RouteError {...props} className="min-h-svh" />
```
nrjdalal: should be at body like app layout? what do you think?

web/next/src/components/route-loading.tsx
nrjdalal, line 1: should be like at components/zeroui or something? like gotta audit more things to group together other than this too

PR #603, opened 2026-07-01
feat(cli): make init's DB provisioning interactive, refresh the demo

nrjdalal:
## What

`init` now offers to provision a local Postgres for you, runs quietly (only high-level steps), and the demo is refreshed to show the real flow.

## init behavior

- **Always asks** `Provision a local Postgres database now?` (defaults to yes when Docker is running, no when it isn't) — previously the offer only appeared when Docker was already up.
- **On yes it does everything**: `pglaunch` → writes `POSTGRES_URL` into `.env` → migrates. The user never runs those by hand.
- **`--yes` provisions too** when Docker is up (non-interactive takes the prompt's default); `--db` still forces it and surfaces the Docker-down guidance.
- **No auth prompt** — social login is a passive note (add GitHub/Google OAuth creds to `.env` anytime); everything else has a working default, so a fresh scaffold runs as-is.
- **Quiet output**: dropped the `.env` narration and run `db:migrate` captured, so provisioning no longer spews the turbo/drizzle build logs. init prints just the steps (fetch, rebrand, install, provision) + the ready summary.
- `getting-started/setup.mdx` synced.

## Demo

- Terminal height `480` → `640`; harness drops the Docker stub (Docker must be up when recording) and clears any prior `pglaunch` container so each run provisions fresh.
- Re-recorded showing the real flow — the whole thing now fits one clean frame. `gifsicle --lossy=30` in the render step (documented in the tape header); with the quiet output the GIF is ~90KB.

Verified: `tsc` clean; real `--yes` and `--yes --db` runs both provisioned + migrated + wrote `POSTGRES_URL`; recorded GIF shows the DB prompt → provisioning → `✓ ready`.

## Before / after

| Before (#529) | After |
| --- | --- |
| before | after |

## Summary by CodeRabbit

* **New Features**
  * The setup flow now better handles database provisioning by automatically using local PostgreSQL when available, while giving clearer guidance when Docker isn’t running.
  * Post-install messaging has been improved to show the right next steps for local or hosted database setups.

* **Bug Fixes**
  * Cleaned up setup behavior to avoid leftover database containers interfering with demo or initialization runs.

* **Documentation**
  * Expanded setup instructions to better explain what the initialization command does and when database setup happens.


web/next/src/app/page.tsx
```diff
@@ -191,22 +191,12 @@ export default async function Home() {
 // Fully typed { data, error }, TypeScript knows exactly what you're getting!
 const { data, error } = await unwrap(apiClient.health.$get())`
 
-  const bashCode = `# Clone the template
-bunx gitpick ${site.social.github}/tree/main
-cd zerostarter
+  const bashCode = `# Scaffold a fresh product in one command — fetches, rebrands,
+# installs, provisions a local Postgres, and migrates:
+zerostarter init
```
nrjdalal: bunx zerostarter init

PR #625, opened 2026-07-03
fix(cli): reuse a running local Postgres on init and wait for readiness

nrjdalal:
## What

`npx zerostarter@latest init` aborted the database step on a re-run:

```
Provisioning a local database with pglaunch and migrating ...
  Database setup failed; set POSTGRES_URL in .env yourself.
  Error:
  - Specify a different name with the `-n` flag, e.g. -n my-project.
  - Or use `-c` flag to start another container with a similar name.
```

pglaunch names its container after the directory (`<dir>-<rand4>`) and, on a
second run, refuses a similar-named container unless `-c` is passed, exiting
non-zero. The old `provisionDatabase` let that failure propagate.

## Fix

`provisionDatabase` (`packages/cli/src/db.ts`) now:

1. **Reuses the already-running Postgres.** On a name collision pglaunch still
   prints the existing container's `POSTGRES_URL`; we parse it and point `.env`
   there instead of failing. Re-running `init` is now idempotent.
2. **Waits for readiness before migrating.** pglaunch returns the moment
   `docker run` starts, so migrating immediately raced the Postgres boot and
   failed even on a clean first launch. We now poll `pg_isready` in the
   container (postgres:alpine ships it) before running `db:migrate`.
3. **Falls back to a fresh container** (`pglaunch -k -c`) only if the reused
   database rejects the migrations.

## Verification

- New unit tests for `parseLaunch` cover the fresh-launch line, the
  ANSI-colored collision line, and the no-URL case. Full CLI suite: 65 pass.
- End-to-end against Docker: a fresh launch now waits out the boot and migrates
  cleanly (previously exit 1), and a re-run reuses the same container (one
  container, same URL, idempotent migrate). `lint`, `check-types`, `build` green.

Bumps the CLI to **0.0.16** (manual bump so `cli-release` publishes on merge to main).

Docs: `setup.mdx` now notes that re-running `init` reuses an already-running local Postgres.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved local database setup so re-running initialization can reuse an existing PostgreSQL container when available.
  * Added a safer fallback to create a fresh container if the first attempt doesn’t work, helping setup complete more reliably.

* **Documentation**
  * Clarified the getting-started guide to explain how local database provisioning behaves during setup.

* **Chores**
  * Bumped the CLI package version.

### DIFF
--- a/packages/cli/package.json
+++ b/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "zerostarter",
-  "version": "0.0.15",
+  "version": "0.0.16",
   "description": "Go from zero to a production-ready SaaS, rebranded and ready to ship.",
   "keywords": [
     "cli",
--- a/packages/cli/src/db.ts
+++ b/packages/cli/src/db.ts
@@ -72,16 +72,20 @@ export const hasPostgresUrl = (dir: string): boolean => {
   return exists(envPath) && getEnvVar(envPath, "POSTGRES_URL") !== ""
 }
 
-// A launched (or reused) local Postgres: the connection URL and its Docker container name.
-type Launch = { url: string; container: string }
+// A launched (or reused) local Postgres: the connection URL, its Docker container name, and whether pglaunch reused an already-running container (reused) rather than starting a new one.
+type Launch = { url: string; container: string; reused: boolean }
 
-// Pull the connection URL and container name out of pglaunch's output. pglaunch prints both whether it just started a container (`... name "<name> :<port>" started ...`) or, on a name collision, reports an already-running one (`... similar name "<name>" running ...`); the URL line is ANSI-colored. Returns null when no URL is present.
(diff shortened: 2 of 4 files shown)


nrjdalal:
Addressed the reuse-path hardening from review in 263e8e4a.

Rather than classifying drizzle-kit's (opaque) exit, `parseLaunch` now reports whether pglaunch **reused** an already-running container or **started** a fresh one. `provisionDatabase` only falls back to a clean container when it just created that container (empty, nothing to lose). If a **reused** database rejects the migrations, the error now surfaces instead of silently repointing `.env` at a fresh, empty container and orphaning data.

Verified end-to-end: with a failing `db:migrate` against a reused container, the run exits non-zero, `.env` still points at the reused DB, and no replacement container is created. Fresh-launch and reuse happy paths unchanged (one container, idempotent migrate). 8 `parseLaunch` unit tests (adds the `reused` field + an ambiguous-output-defaults-to-reuse case).

PR #659, opened 2026-07-06
fix: preserve shadcn --sidebar override, refresh CLI demo timing, add CLI README

nrjdalal:
Bundled fixes (one PR per request). Four commits, three concerns:

### 1. shadcn sync: preserve the `--sidebar` globals.css override

`shadcn init` regenerates the whole `globals.css` from the upstream template, and `shadcn-customize.ts`'s `patchGlobals()` only re-applied the `--font-sans` override. The `--sidebar: var(--background)` override (flush sidebars with the background, from #566) was never encoded, so every `bun run shadcn:update` reverted both `--sidebar` lines to shadcn's defaults, silently clobbering the customization. `patchGlobals()` now re-applies the sidebar override (both `:root` + `.dark`), guarded on finding exactly 2 so an upstream shape change fails the sync loudly. Skill doc updated. Verified: sync now reproduces `globals.css` with an empty diff.

### 2. CLI demo GIF: refreshed for v0.0.23 and fixed the timing

- Re-recorded `.github/assets/cli.gif` against the current CLI. The demo was last recorded at #643; #645's flow polish (titled steps + muted detail tails, streamed migrate output) had drifted from the committed GIF.
- Fixed the recording timing: the blind `Sleep 16s` answered the Postgres prompt long after it appeared (install is usually cached), leaving dead air. The tape now uses `Wait+Screen` keyed to the prompt and the outro, so the `Enter` fires ~1.5s after the prompt renders. Runtime ~29s → ~17s, interactive choice still clearly visible.

### 3. CLI package README

Added `packages/cli/README.md` documenting `init` / `reinit` / `sync` (options mirrored from each command's `--help`), quick start, and requirements, so the npm page for `zerostarter` isn't bare. Uses an absolute raw-GitHub GIF URL so it renders on npm.

---

See the thread comments for before/after GIFs and verification detail.


nrjdalal:
### Also in this PR: refreshed the CLI init demo GIF

The README demo (`.github/assets/cli.gif`) was last recorded at #643. #645 (`error legibility, friendly flags, and flow polish`) changed the init output but the GIF was never refreshed, so it was stale. Re-recorded against the current CLI (`v0.0.23`) with the existing `cli.tape` (no tape edits needed, one prompt → one `Enter`).

What changed in the output the new capture reflects:
- Install is now a titled step **Installed dependencies** with a muted detail tail (`1715 packages installed` / `Removed: 1`), instead of a single inline line.
- The DB step is now **Provisioned a local Postgres database** streaming `bun run db:migrate` output as the muted tail, instead of `Provision a local Postgres database now? Yes` + `2 migrations applied`.

| Before (#643-era) | After (v0.0.23) |
| --- | --- |
| before | after |

Verified by exploding the recorded GIF to frames: the flow completes end-to-end (fetch → rebrand → install → Postgres provisioned → "acme is ready" → outro), the prompt is answered, and it ends cleanly at the shell. Optimized with `gifsicle -O3 --lossy=30` per the tape's documented pipeline (207 KB).

_Note (not addressed here): the migrate step surfaces drizzle-kit's raw preamble (`No config path provided…`, `Reading config file…`, `Using 'postgres' driver…`) as its detail tail, which reads noisier than the old `N migrations applied` summary. That's current `runTail` behavior from #645, a separate CLI-UX call if you want it quieted._

nrjdalal:
### Follow-ups: demo timing fix + CLI package README

**1. Demo timing (`fix(cli): wait for the Postgres prompt…`)**

The first re-record answered the Postgres prompt on a blind `Sleep 16s`. Since `bun install` is usually cached and finishes in a couple of seconds, the prompt sat answered-late, i.e. dead air before the "Yes". Replaced the fixed sleeps with `Wait+Screen` keyed to the prompt text and the outro, so the `Enter` fires ~1.5s after the prompt actually renders (regardless of install/cache speed) and the recording ends right after the final note.

| Before (blind 16s sleep) | After (wait-based) |
| --- | --- |
| before | after |

Runtime dropped from ~29s to ~17s, and the interactive `Provision a local Postgres database? ● Yes / ○ No` choice is still clearly visible (verified by exploding the optimized GIF to a frame contact-sheet: the prompt holds across ~1.5s of frames, then provisioning proceeds to "acme is ready").

**2. CLI package README (`docs(cli): add package README…`)**

`packages/cli` had no README, so the npm page for `zerostarter` was bare. Added one documenting `init` / `reinit` / `sync` with their options (mirrored from each command's `--help`), quick start, and requirements. npm publishes `README.md` regardless of the `files` allowlist, so the demo GIF uses an absolute `raw.githubusercontent.com/.../canary/...` URL (verified `200`) to render on the npm page.

PR #660, opened 2026-07-06
feat(env): scoped SKIP_ENV_VALIDATION_SERVER and _CLIENT flags

nrjdalal:
Adds scoped env-validation skip flags so a build can bypass just the server or just the client section, alongside the existing all-or-nothing `SKIP_ENV_VALIDATION`.

## Why

`SKIP_ENV_VALIDATION=true` skips *everything* — the Docker/CI escape hatch. But some builds legitimately have one set of vars and not the other (e.g. a web build that must still guarantee the inlined `NEXT_PUBLIC_*` client vars but has no server secrets). This adds a scalpel without removing the hammer.

## Flags & precedence

| Flag set | server section | client section |
| --- | --- | --- |
| (none) | validated | validated |
| `SKIP_ENV_VALIDATION` | skipped | skipped |
| `SKIP_ENV_VALIDATION_SERVER` | skipped | validated |
| `SKIP_ENV_VALIDATION_CLIENT` | validated | skipped |

Base implies both: a section is skipped when `SKIP_ENV_VALIDATION` **or** its scoped flag is set. The naming is a suffix family (`SKIP_ENV_VALIDATION`, `..._CLIENT`, `..._SERVER`) so the flags group and sort together.

## Implementation

- New `packages/env/src/lib/skip.ts` computes `skipClientValidation` / `skipServerValidation` from the three vars (base OR scoped).
- Server-only packages (`api-hono`, `auth`, `db`) map `skipValidation` to `skipServerValidation` (`_CLIENT` never applies — they have no client section).
- `web-next` is split into two `createEnv` calls — a `serverEnv` (skipped by `skipServerValidation`) and a client env that `extends: [serverEnv]` (skipped by `skipClientValidation`). `extends` is native to `@t3-oss/env-core` 0.13, so the exported `env` still carries both sets. The `NEXT_PUBLIC_*` polyfill now keys off `skipClientValidation`.
- `turbo.json` `globalEnv` gains the two flags so cache keys track them.
- Docs (`manage/environment.mdx`) document the flags and precedence.

## Testing

- Proved the full truth table above by loading the composed `web-next` env under each flag combination (client-skip polyfills the missing `NEXT_PUBLIC_*` and leaves server enforced; server-skip lets an invalid `INTERNAL_API_URL` through while still enforcing the client vars; base skips both).
- `bunx turbo run build` (7/7) and `check-types` (12/12, CI-style with the skip flag) pass in a fresh worktree.
- Confirmed `SKIP_ENV_VALIDATION_SERVER=true` alone unblocks the server-package builds without a `.env`.

### DIFF
--- a/packages/env/src/api-hono.ts
+++ b/packages/env/src/api-hono.ts
@@ -3,6 +3,7 @@ import { z } from "zod"
 
 import "@/lib/utils"
 import { NODE_ENV } from "@/lib/constants"
+import { skipServerValidation } from "@/lib/skip"
 
 export const env = createEnv({
   server: {
@@ -25,5 +26,5 @@ export const env = createEnv({
     HONO_TRUSTED_ORIGINS: process.env.HONO_TRUSTED_ORIGINS,
   },
   emptyStringAsUndefined: true,
-  skipValidation: process.env.SKIP_ENV_VALIDATION === "true",
+  skipValidation: skipServerValidation,
 })
--- a/packages/env/src/auth.ts
+++ b/packages/env/src/auth.ts
@@ -3,6 +3,7 @@ import { z } from "zod"
 
 import "@/lib/utils"
 import { NODE_ENV } from "@/lib/constants"
+import { skipServerValidation } from "@/lib/skip"
 
 export const env = createEnv({
   server: {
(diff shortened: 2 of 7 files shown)


nrjdalal:
Superseded: the scoped SKIP_ENV_VALIDATION_SERVER/_CLIENT split was the wrong abstraction. The real issue is that the frontend build validates server env at all. Replacing with a fix that removes build-time env validation from the package tsdown configs, so each env file is validated at runtime by the package that owns it, and the frontend build only needs web-next's own vars.

PR #661, opened 2026-07-06
feat: stream system health over a WebSocket (live status demo)

nrjdalal:
## Summary

Introduces WebSocket support to the stack, using the health endpoint as the live demo. The landing page's "All systems are operational" badge shows a live pulse from a WebSocket, over a REST poll that keeps it honest.

## What changed

**Backend** (`api/hono/src/index.ts`)
- New `/api/health/ws` route: a `GET` upgraded with `upgradeWebSocket` from `hono/bun`. Pushes a health snapshot on connect, then a heartbeat every 5s; the interval is cleared in `onClose` (Bun's WS adapter surfaces every disconnect as `close`).
- The shared `websocket` handler is registered on the Bun server export next to `fetch`.
- Documented with `describeRoute`, so it lists in the Scalar reference at `/api/docs` as a `101` upgrade. OpenAPI can't schema-type WS frames, so the frame shape is described in prose.
- The REST `/api/health` endpoint is untouched.

**Frontend** (`components/marketing/api-status.tsx`)
- REST is the always-honest baseline: `apiClient.health.$get()` is polled (TanStack Query) whenever no frame is live, and the socket (`apiClient.health.ws.$ws()`) overlays a live pulse while it streams. The badge prefers a live frame and otherwise reflects REST.
- The socket reconnects with capped exponential backoff (reset on a frame and on tab refocus), a rolling 12s heartbeat watchdog catches a half-open connection, and a socket-identity guard stops a superseded socket's late events from disturbing a newer one. Frames aren't RPC-typed, so it parses defensively and reads `message`.
- Moved the status badge below the Quickstart window in the hero.

**Docs**
- "Live data over WebSockets" section in the type-safe API guide, and a "WebSocket routes" section in the `api-endpoint` skill (including a note that CORS doesn't gate a WS handshake).

## Verification

`check-types` (12/12), `oxlint`, `oxfmt --check`, and docs strict-validation all pass. Verified end-to-end in a real browser.

- WebSocket probe: snapshot on connect, heartbeat at 5s, clean close.
- `/api/openapi.json` lists `/api/health/ws` (GET -> `101`) under the **System** tag; `/api/docs` renders it.
- Live: green dot with `animate-pulse`.
- WS unavailable but API up (serverless-like): green **static** dot via the REST baseline, no pulse.
- Outage: killing the API flips the badge from green to `connecting`, then REST surfaces it as `down`; on restart the pulse recovers.

### Before / after (hero)

| Before (poll, badge above Quickstart) | After (WS live, badge below Quickstart) |
| --- | --- |
| before | after |

## Deployment

WebSockets hold under Bun (local + Docker/self-hosted) but not on standard Vercel serverless functions. Since REST is the baseline, a serverless deploy simply shows a normal green (non-pulsing) badge; the live pulse appears wherever the socket holds.

### DIFF
--- a/api/hono/src/index.ts
+++ b/api/hono/src/index.ts
@@ -16,14 +16,6 @@ import { agentsRouter, authRouter, v1Router, waitlistRouter } from "@/routers"
 
 const BUILD_VERSION = getBuildVersion()
 
-// Live payload pushed over /api/health/ws: the REST health data plus an ISO timestamp per heartbeat.
-export type HealthEvent = {
-  message: string
-  version: string
-  environment: typeof env.NODE_ENV
-  timestamp: string
-}
-
 const app = new Hono()
 
 app.use(
@@ -100,18 +92,41 @@ const { data, error } = await unwrap(apiClient.health.$get())`,
   )
   .get(
     "/health/ws",
+    describeRoute({
+      tags: ["System"],
+      description:
+        "Live system health over a WebSocket (Bun). On connect the server sends a snapshot, then a heartbeat every 5s. Each frame is JSON: { message, version, environment, timestamp }.",
+      ...({
+        "x-codeSamples": [
+          {
+            lang: "typescript",
+            label: "hono/client",
+            source: `import { apiClient } from "@/lib/api/client"
+
+const socket = apiClient.health.ws.$ws()
+socket.addEventListener("message", (event) => {
+  const health = JSON.parse(event.data)
+})`,
+          },
+        ],
+      } as object),
+      responses: {
+        101: { description: "Switching Protocols: the WebSocket handshake succeeded." },
+      },
+    }),
     upgradeWebSocket(() => {
       let heartbeat: ReturnType<typeof setInterval> | null = null
-      const snapshot = (): HealthEvent => ({
-        message: "ok",
-        version: BUILD_VERSION,
-        environment: env.NODE_ENV,
-        timestamp: new Date().toISOString(),
-      })
+      const snapshot = () =>
+        JSON.stringify({
+          message: "ok",
+          version: BUILD_VERSION,
+          environment: env.NODE_ENV,
+          timestamp: new Date().toISOString(),
+        })
       return {
         onOpen(_event, ws) {
-          ws.send(JSON.stringify(snapshot()))
-          heartbeat = setInterval(() => ws.send(JSON.stringify(snapshot())), 5000)
+          ws.send(snapshot())
+          heartbeat = setInterval(() => ws.send(snapshot()), 5000)
         },
         onClose() {
           if (heartbeat) clearInterval(heartbeat)
--- a/api/hono/src/index.ts
+++ b/api/hono/src/index.ts
@@ -4,6 +4,7 @@ import { env } from "@packages/env/api-hono"
 import { Scalar } from "@scalar/hono-api-reference"
 import { Hono } from "hono"
 import { describeRoute, openAPIRouteHandler, resolver } from "hono-openapi"
+import { upgradeWebSocket, websocket } from "hono/bun"
 import { cors } from "hono/cors"
 import { HTTPException } from "hono/http-exception"
 import { logger } from "hono/logger"
@@ -14,6 +15,14 @@ import { rateLimiterMiddleware } from "@/middlewares"
 import { agentsRouter, authRouter, v1Router, waitlistRouter } from "@/routers"
 
 const BUILD_VERSION = getBuildVersion()
+
+// Live payload pushed over /api/health/ws: the REST health data plus an ISO timestamp per heartbeat.
+export type HealthEvent = {
+  message: string
+  version: string
+  environment: typeof env.NODE_ENV
+  timestamp: string
+}
 
 const app = new Hono()
 
@@ -89,6 +98,27 @@ const { data, error } = await unwrap(apiClient.health.$get())`,
       return c.json({ data })
     },
   )
+  .get(
+    "/health/ws",
+    upgradeWebSocket(() => {
+      let heartbeat: ReturnType<typeof setInterval> | null = null
+      const snapshot = (): HealthEvent => ({
+        message: "ok",
+        version: BUILD_VERSION,
+        environment: env.NODE_ENV,
+        timestamp: new Date().toISOString(),
+      })
+      return {
+        onOpen(_event, ws) {
+          ws.send(JSON.stringify(snapshot()))
+          heartbeat = setInterval(() => ws.send(JSON.stringify(snapshot())), 5000)
+        },
+        onClose() {
+          if (heartbeat) clearInterval(heartbeat)
+        },
+      }
+    }),
+  )
   .route("/agents", agentsRouter)
   .route("/auth", authRouter)
   .route("/v1", v1Router)
@@ -130,4 +160,5 @@ export type { ErrorCode } from "@/lib/error"
 export default {
   port: env.HONO_PORT,
   fetch: app.fetch,
+  websocket,
 }
--- a/api/hono/src/index.ts
+++ b/api/hono/src/index.ts
@@ -116,10 +116,6 @@ socket.addEventListener("message", (event) => {
     }),
     upgradeWebSocket(() => {
       let heartbeat: ReturnType<typeof setInterval> | null = null
-      const stop = () => {
-        if (heartbeat) clearInterval(heartbeat)
-        heartbeat = null
-      }
       const snapshot = () =>
         JSON.stringify({
           message: "ok",
@@ -132,11 +128,9 @@ socket.addEventListener("message", (event) => {
           ws.send(snapshot())
           heartbeat = setInterval(() => ws.send(snapshot()), 5000)
         },
+        // Bun's WS adapter surfaces every disconnect as close (it has no error event), so this covers all cleanup.
         onClose() {
-          stop()
-        },
-        onError() {
-          stop()
+          if (heartbeat) clearInterval(heartbeat)
         },
       }
     }),
--- a/api/hono/src/index.ts
+++ b/api/hono/src/index.ts
@@ -116,6 +116,10 @@ socket.addEventListener("message", (event) => {
     }),
     upgradeWebSocket(() => {
       let heartbeat: ReturnType<typeof setInterval> | null = null
+      const stop = () => {
+        if (heartbeat) clearInterval(heartbeat)
+        heartbeat = null
+      }
       const snapshot = () =>
         JSON.stringify({
           message: "ok",
@@ -129,7 +133,10 @@ socket.addEventListener("message", (event) => {
           heartbeat = setInterval(() => ws.send(snapshot()), 5000)
         },
         onClose() {
-          if (heartbeat) clearInterval(heartbeat)
+          stop()
+        },
+        onError() {
+          stop()
         },
       }
     }),
--- a/web/next/src/components/marketing/api-status.tsx
+++ b/web/next/src/components/marketing/api-status.tsx
@@ -1,11 +1,11 @@
 "use client"
 
+import { useQuery } from "@tanstack/react-query"
 import { useEffect, useState } from "react"
 
 import { Badge } from "@/components/ui/badge"
-import { apiClient } from "@/lib/api/client"
-
-type ConnectionState = "connecting" | "live" | "down"
+import { apiClient, unwrap } from "@/lib/api/client"
+import { cn } from "@/lib/utils"
 
 // Frames aren't RPC-typed (Hono types the route, not the payload), so read the one field we need defensively.
 function isOperational(raw: unknown): boolean {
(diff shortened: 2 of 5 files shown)


api/hono/src/index.ts
```diff
@@ -15,6 +16,14 @@ import { agentsRouter, authRouter, v1Router, waitlistRouter } from "@/routers"
 
 const BUILD_VERSION = getBuildVersion()
 
+// Live payload pushed over /api/health/ws: the REST health data plus an ISO timestamp per heartbeat.
+export type HealthEvent = {
```
nrjdalal: we can inline this for now? if single use?

web/next/src/components/marketing/api-status.tsx
```diff
@@ -1,22 +1,56 @@
 "use client"
 
-import { useQuery } from "@tanstack/react-query"
+import type { HealthEvent } from "@api/hono"
```
nrjdalal: rpc doesn't work? why do we have to do this?

api/hono/src/index.ts
nrjdalal: Done. Dropped the exported `HealthEvent` type and inlined the payload: `snapshot()` returns the JSON string directly now. It was actually used in two places (the frontend imported it too), but per the next comment RPC can't type WS frames anyway, so the shared type bought nothing.

web/next/src/components/marketing/api-status.tsx
nrjdalal: Checked Hono's own types to be sure. RPC types the route **path**, not the frames:

- `WSEvents.onMessage` hands you `MessageEvent<string | Blob | ArrayBufferLike>`
- `ws.send()` takes a raw `string | ArrayBuffer | Uint8Array`
- the client is `$ws: (args?) => WebSocket` — a plain `WebSocket`, `data` is untyped

There's no generic to flow a payload type end to end. So `apiClient.health.ws.$ws()` is typed (the path/URL is real RPC), but the messages on the wire aren't. Removed the manual import; the component only needs `message`, so it parses defensively and checks that one field. Typing frames would mean hand-maintaining a shared type RPC can't derive, which is the surface flagged in the other comment, so I left it off.

api/hono/src/index.ts
```diff
@@ -89,6 +98,27 @@ const { data, error } = await unwrap(apiClient.health.$get())`,
...
+        onOpen(_event, ws) {
+          ws.send(JSON.stringify(snapshot()))
+          heartbeat = setInterval(() => ws.send(JSON.stringify(snapshot())), 5000)
+        },
+        onClose() {
+          if (heartbeat) clearInterval(heartbeat)
+        },
+      }
+    }),
+  )
```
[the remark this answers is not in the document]
nrjdalal: Good catch, fixed. Added an `onError` handler and a shared `stop()` that clears the interval and nulls the handle, called from both `onClose` and `onError`, so a failed socket can't leak the 5s heartbeat.

nrjdalal:
Thanks for the thorough pass. Both observations addressed in the two latest commits (`5affe47`, `bdc4d341`):

**1. Opened-but-silent socket (`api-status.tsx`).** Reworked the component so the socket is progressive enhancement: it drives the live pulse when frames arrive, but a 4s deadline (plus `close`/`error`) now flips it to polling REST `/api/health` via TanStack Query. So a handshake that succeeds but never delivers a frame no longer hangs in the invisible `connecting` state, it resolves to real status.

**2. Serverless red-badge default.** Same change resolves this: on a deploy that can't hold a socket, the badge falls back to a green (non-pulsing) REST-polled state instead of sitting red and retrying every 3s. The old fixed-interval reconnect is gone, once it falls back it stays on the poll for the session.

Verified both in a browser: live path shows the pulsing dot; disabling the WS route so `$ws()` can't upgrade shows the badge fall back to a green static dot (not invisible, not red). The `describeRoute` also lists the route in the Scalar reference now.

nrjdalal:
Good catch on the one-way fallback, addressed in `456fe11a` since this is the WS *demo* page and a blip shouldn't silently end the demo.

The reconnect now distinguishes two cases so it stays resilient without reintroducing the aggressive loop the first pass flagged:

- **A socket that had gone live and drops** (wifi blip, server restart, proxy idle-timeout) retries a few times (3s apart, capped) and resumes the pulse; it also re-probes on `visibilitychange` when the tab returns.
- **A socket that never connected** (serverless, buffered upgrade) still falls straight to REST, so there's no background retry loop and no invisible gap.

A recovered live frame flips `useRest` back off, so the pulse returns and the REST poll disables itself.

Verified all three paths in a browser:
- Live: pulsing dot.
- Never-live (route disabled before load): green **static** dot immediately, not invisible.
- Transient drop (disable the route while live, restore it within the retry window): the pulse **recovers on its own**, no reload.

Docs/skill updated to describe the retry. Thanks for the thorough passes.

nrjdalal:
Fixed in `3b64e0d`. `onVisible` now clears any pending `retry`/`deadline` before reconnecting, so `connect()` can never run while a queued reconnect is still in flight, no second overlapping socket, no orphan, and the stale timer can't fire against the new connection.

Trace matched yours exactly: the gap was `socket === null` during the 3s backoff, which `onVisible`'s `socket !== null` guard doesn't cover. Since `onVisible` only runs when `socket` is null (and `onDrop` has already cleared `deadline`), the only thing `clearTimers()` cancels there is the pending `retry`, then `connect()` opens a single socket.

Happy paths re-verified (live pulse, never-live REST fallback); it's a one-line guard so nothing else moved.

nrjdalal:
Both addressed.

**1. Half-open watchdog (`0746f09`).** You're right that the deadline was a first-frame guard only, so the heartbeat's liveness signal was thrown away after frame #1. It's now a rolling watchdog: each `message` re-arms `deadline` with a 12s timeout (> the server's 5s heartbeat, ~2 missed beats of slack). A heartbeat gap on a half-open socket now trips `onDrop` → retry, and if frames don't resume it settles to REST, so the badge can't sit on a stale "operational" pulse for a hung server.

A note on verifying this one: a genuine half-open socket (TCP up, server silent, no FIN) isn't reproducible with dev-server edits, since any server change hot-reloads and closes the socket cleanly, which is the `close`-event path already covered. So I verified the re-arm by construction and confirmed it doesn't regress normal streaming: the pulse holds steady across multiple 5s heartbeats without the 12s watchdog tripping.

**2. Doc double-encode (`68bc8ca`).** Fixed. The snippet now sends the object inline (`ws.send(JSON.stringify({ message: "ok", version, environment, timestamp }))`), so it no longer implies `snapshot()` returns an object when the shipped helper returns a string.

nrjdalal:
All three fixed in `81cf2bc`.

**1. Stale pulse during the retry window.** You're right, the pulse conflated "last-known health" with "currently streaming." Split the state: `wsHealth` (last status seen) drives the badge color; `wsConnected` (set false in `onDrop`, true on each frame) drives the pulse, and the render does `live = wsConnected`. So a dropped-but-reconnecting socket now shows the last-known status as a **static** dot, and the pulse is genuinely socket-only. Verified: with no live socket the dot is `bg-success ... rounded-full` (no `animate-pulse`); once a frame arrives it's `... animate-pulse`.

**2. `onVisible` starving the REST fallback.** Dropped the `retries = 0` from `onVisible`. The frame handler already resets `retries` on a successful frame, so a recovered socket still gets a fresh budget, while a socket that keeps connecting-then-dropping (even with visibility toggling around each failure) now advances `retries` to `RETRY_LIMIT` and reaches `setUseRest(true)`. `clearTimers()` before `connect()` stays, so no overlapping socket.

**3. Multi-line `//` comments.** Collapsed the three offending blocks to single lines per the root `CLAUDE.md` rule. `oxfmt`/`oxlint` clean.

`check-types` 12/12, lint, format all green. (No doc change needed, the "pulse is socket-only" claim is now literally true.)

nrjdalal:
All three fixed (`8bf3c07`, `7ca2259`), and I verified finding 1 end-to-end by killing a live API.

**1. Stale green during the reconnect window.** Right, gating only the pulse wasn't enough, the green status itself was unwarranted while disconnected. `operational`/`down` are now gated on `wsConnected`, so a dropped socket renders `connecting` (invisible) until a frame or REST resolves it. Verified live: with a socket streaming (green + pulse) I killed the API; the badge went straight to the invisible `connecting` state (class flips `bg-success...` -> `invisible...`, no stale green), then after the retries exhausted REST surfaced the outage as `down` ("Systems are facing issues", `bg-destructive`). It no longer asserts "operational" for an unreachable server.

**2. `onVisible` retry budget vs. the prior round's infinite-retry fix.** These two notes pull opposite ways on the same line, so I reconciled them: `onVisible` now guards on `retry === null`, so it only re-probes when *settled on REST* (no live socket and no pending retry) and resets `retries = 0` there. During an active retry burst `retry` is non-null, so `onVisible` bails, which keeps a flapping socket falling through to REST (closes the earlier infinite-loop) while a genuine refocus-from-REST still gets a full budget to recover (closes this one).

**3. Dead `onError`.** Confirmed, `hono/bun`'s adapter maps only `open`/`close`/`message` (Bun has no WS error event), so `onError` never fired. Removed it plus the single-use `stop()` helper; `onClose` covers all cleanup, with a one-line comment saying why.

On the non-blocking OpenAPI note: leaving the `429`/`500` on the upgrade route as-is. The `429` is actually reachable (the rate limiter runs on the upgrade GET) and `500` is the global default; suppressing them per-route adds machinery for a route OpenAPI can't fully model anyway. Happy to strip them if you'd rather.

Green: types 12/12, lint, format, docs strict.

nrjdalal:
Findings 1-3 were each a symptom of the same thing: a hand-rolled retry/fallback state machine that's produced a fresh edge case almost every review round (downgrade-on-slow-frame, superseded-socket race, retry-budget accounting). Rather than patch each (they interact), I collapsed it to a simpler model in `c9193a1`, and did the doc fixes in `bde71c0`.

**New model: REST is the always-honest baseline; the socket is a pure pulse overlay.**
- REST `/api/health` polls whenever no frame is live (`enabled: wsFrame === null`, 30s). The socket only sets `wsFrame` and the badge prefers it when present.
- The socket reconnects with capped exponential backoff (3s -> 30s), reset on any frame and on refocus.

That structurally removes all three findings:

**1. Slow first frame downgrading the session.** Gone, there's no `FRAME_DEADLINE` anymore. A slow or silent socket just means REST carries the badge until a frame arrives; nothing closes a working-but-slow socket, and the pulse appears whenever the frame lands.

**2. Superseded socket tearing down a live one.** Fixed with an identity guard: every handler closes over its own `ws` and bails unless `socket === ws`, so a late `close`/`error`/watchdog from an old socket can't touch the current one.

**3. `wasLive` retry ladder burning ~9s on refocus.** Gone, no `wasLive`, no ladder. A dead endpoint just backs off to a 30s re-probe while REST keeps the badge honest; refocus resets the backoff for an immediate try.

**4. Skill said `access.tsx` "polls".** It uses `staleTime: Infinity` with no `refetchInterval` (a cached one-shot). Reworded to "reading REST data."

**5. `101` route shows `429`/`500` JSON in Scalar.** Leaving as-is: the `429` is genuinely reachable (the rate limiter runs on the upgrade GET) and `500` is the global default, and there's no clean per-route way to drop the shared `defaultOptions.GET` responses. Documenting a WS upgrade in an HTTP spec has this seam inherently. Easy to revisit if you'd rather pull the route from OpenAPI.

**6. WS handshake not CORS-gated; `$ws()` drops credentials.** Added a skill note: CORS doesn't apply to a WS handshake and `$ws()` sends no credentials, so a sensitive/authed route must check `Origin` (or a token) in the handler. `/api/health/ws` is public data, so it stays ungated, consistent with that guidance.

Verified: types 12/12, lint, format, docs strict; and in a browser, live socket -> green + pulse, WS unavailable with the API up (serverless-like) -> green static via REST, recovery -> pulse returns.

nrjdalal:
**Finding 1 (orphaned half-open socket) — fixed in `8eb5728`.** Good catch, that was a real leak. `drop()` now calls `ws.close()` at the top: on the watchdog path (socket still `OPEN`, heartbeats stopped, no close event) it actually closes the orphan so the server's `onClose` fires and clears its 5s interval instead of streaming to a dead peer; on the `close`/`error` paths it's a no-op since the socket is already closing. Verified the live pulse and reconnect are unaffected.

**Finding 2 (OpenAPI `429`/`500` on the `101`) — keeping it, consciously.** This is a real product tradeoff, not an oversight. The `429` is genuinely reachable (the rate limiter runs on the upgrade GET), and `500` is the global default; the only ways to drop them are to pull `/health/ws` from the spec entirely (losing the discoverability + `$ws()` code sample in Scalar that this PR added on request) or to stop using the shared `defaultOptions.GET` and repeat the error responses on every route by hand. I'd rather keep the endpoint discoverable and accept that documenting a WS upgrade inside an HTTP spec has this seam. If you'd prefer a clean spec over the Scalar listing, I'll pull it out, just say which way you want it.

Green: types 12/12, lint, format, docs strict.

nrjdalal:
Filed the cosmetic OpenAPI note (429/500 shown on the `/api/health/ws` `101` route in Scalar) as #664 to track as a follow-up, keeping it out of this PR's scope. Decision there is between leaving it (discoverability) and opting the WS route out of the shared `defaultOptions.GET` responses.

PR #662, opened 2026-07-06
refactor(env): validate server env at runtime, not at build time

nrjdalal:
Makes env a **runtime** concern instead of a build-time one, so building a package never requires another package's variables. Concretely: a Next.js frontend build no longer demands `POSTGRES_URL` (or `BETTER_AUTH_SECRET`, etc.) — each env file is validated at runtime by the package that owns it.

## The problem

The web build is `turbo build --filter=@web/next`, which builds the whole dependency graph including `@packages/db`, `@packages/auth`, `@api/hono`. Each of those `tsdown.config.ts` did `import { env } from "@packages/env/<pkg>"`, and t3-env's `createEnv` validates eagerly at import — so merely *loading the build config* validated that package's server env. Building the frontend therefore required every server secret, none of which it uses. (This is what caused the recent web deploy failure after those "unused" vars were removed from Vercel.)

## The fix

- `definePackageConfig` (`packages/config/tsdown.ts`) no longer reads env — no `build:prepare` validation hook. Building a package validates nothing.
- `db` and `api-hono` drop env from their tsdown configs entirely.
- `@packages/env/auth` is now **lazy** (`lazyEnv` in `packages/env/src/lib/lazy.ts`): `createEnv` runs on first property access, not at import. The auth server validates on first access at runtime; the build never does.
- Each env still validates at runtime (verified: accessing `db`/`auth` env with a required var missing throws; `next build` still validates the `NEXT_PUBLIC_*` client vars for inlining).

## Why not just annotate the type (the "clean" alternative)

`@packages/auth`'s `betterAuth(...)` export has an inferred type that references a zod-v4 internal (`$strip`), which isn't dts-portable. Annotating it (`auth: Auth` / `as unknown as Auth`) makes dts build but **widens the type and breaks the app** — `activeOrganizationId` (org plugin) and `role` (admin plugin) disappear from the inferred `Session`/`User`, which `(protected)/layout.tsx` and `console.ts` depend on. There is no self-contained annotation that keeps the plugin types. So auth's tsdown config keeps a `@packages/env` import purely to hold a zod reference in its tsconfig program (it `include: ["**/*.ts"]`) so tsgo can emit the dts — documented inline. The env is lazy, so this import doesn't validate at build.

## Verification

- `bunx turbo build check-types` — **14/14 pass with all server env UNSET and no `SKIP_ENV_VALIDATION`** (only the client vars a web build needs). This is the goal.
- Runtime validation preserved: `env.POSTGRES_URL` / `env.BETTER_AUTH_SECRET` throw on access when unset, return the value when set.
- `oxlint` + `oxfmt` clean.

### DIFF
--- a/api/hono/tsdown.config.ts
+++ b/api/hono/tsdown.config.ts
@@ -1,10 +1,5 @@
 import { definePackageConfig } from "@packages/config/tsdown"
-import { getSafeEnv } from "@packages/env"
-import { env } from "@packages/env/api-hono"
 
 export default definePackageConfig({
-  name: "@api/hono",
-  env,
-  getSafeEnv,
   deps: { alwaysBundle: [/^@packages\//], neverBundle: ["bun"] },
 })
--- a/packages/auth/tsdown.config.ts
+++ b/packages/auth/tsdown.config.ts
@@ -2,5 +2,5 @@ import { definePackageConfig } from "@packages/config/tsdown"
 import { getSafeEnv } from "@packages/env"
 import { env } from "@packages/env/auth"
 
-// `env` and `getSafeEnv` are passed only to keep these imports in the tsconfig program (auth's tsconfig includes this file). Better-auth's inferred `auth` type references a zod internal, and the zod re-exported through `@packages/env` is what keeps that reference nameable so tsgo can emit the dts. `env` is lazy, so importing it here does not validate at build time.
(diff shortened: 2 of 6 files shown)


nrjdalal:
Superseded by a scoped build-time approach: instead of moving validation to runtime, only the server-only vars fall back to dummies under SKIP (validation stays on at build). See the replacement PR.

PR #663, opened 2026-07-06
fix(env): fall back server-only vars under SKIP so the frontend build doesn't require them

nrjdalal:
The web build (`turbo build --filter=@web/next`) compiles the server packages (`@packages/db`, `@packages/auth`, `@api/hono`) for their types, and compiling them validated their env — forcing `POSTGRES_URL`, `BETTER_AUTH_SECRET`, `HONO_*` onto a frontend that never uses those values. This makes the frontend build stop requiring server secrets, **without faking the client vars it ships**.

## Design — two scopes

The web build is a distinct case: it has **no** server secrets but **real** client vars that get inlined into the shipped bundle. A single all-or-nothing `SKIP_ENV_VALIDATION` can't express that (it would fake the client vars too, or force throwaway `NEXT_PUBLIC_*` into every workflow). So the fallback is split into two scopes (`polyfillServer` / `polyfillClient`), both substituting a shape-valid dummy for a *missing* var while `skipValidation` stays off so validation always runs:

| Flag | server vars | client (`NEXT_PUBLIC_*`) | used by |
| --- | --- | --- | --- |
| *(none)* | **required, validated** | **required, validated** | API build, all runtimes |
| `SKIP_ENV_VALIDATION` | dummy fallback | dummy fallback | CI / Docker (no real env) |
| `SKIP_ENV_VALIDATION_SERVER` | dummy fallback | **required, validated** | web build (`web/next/vercel.json`) |

- Server vars (`POSTGRES_URL`, `BETTER_AUTH_SECRET`, `HONO_APP_URL`, `HONO_TRUSTED_ORIGINS`) fall back under **either** flag.
- Client vars fall back under the **base** flag only — so the web build validates the public URLs it inlines and ships. A missing/typo'd `NEXT_PUBLIC_*` on the web deploy now **fails the build** instead of shipping `https://polyfill.url`.
- `NODE_ENV` is never faked (CI/Vercel supply it).

CI keeps the single base flag and declares no per-var values; the web build sets the server-scoped flag. `SKIP_ENV_VALIDATION_SERVER` is registered in `turbo.json` `globalEnv`; docs updated in `manage/environment.mdx`.

## Verification

- Web build (`SKIP_ENV_VALIDATION_SERVER`, real client, no server secrets) → **6/6**.
- Web build (`SKIP_ENV_VALIDATION_SERVER`, **no** client vars) → `@web/next` build **fails** (client stays strict).
- CI (base `SKIP_ENV_VALIDATION`, clean env, no throwaway vars) → `turbo build check-types` **14/14**.
- No flag + missing `POSTGRES_URL` → **fails**; `SKIP` + malformed value → **fails**; runtime import with no flag → **throws** (server still required).
- End-to-end: booted the API on a spare port with the real `.env` → validated at startup, served `401`.
- `oxlint` + `oxfmt` clean.

## Follow-up (ops)

After merge, remove the real server secrets from the **web** Vercel project (the build falls them back via `SKIP_ENV_VALIDATION_SERVER` now). Keep real DB/auth secrets on the API project only.


nrjdalal:
## Review fix — client vars kept strict under `SKIP`

Both findings addressed in 54597491.

**Finding 1 (regression):** correct — setting `SKIP` on the production web build activated `web-next`'s inline `NEXT_PUBLIC_*` fallback, so a missing/typo'd public URL would have shipped `https://polyfill.url` silently. Fixed by **removing the inline client fallback entirely** — public vars are inlined into the shipped bundle, so they are now always required and validated, in every build. `SKIP` only falls back the server-only vars now.

**Finding 2 (duplication):** resolved by the same change — the inline fallback that duplicated `polyfill()` is gone; `polyfill()` is the single fallback path, used only for server vars.

The two CI builds that set `SKIP` (`auto-check-build.yml`, `auto-release.yml`) now pass throwaway public URLs explicitly, since they build the web app without real ones. Docs updated (`manage/environment.mdx`).

**Proof:** `SKIP=true NODE_ENV=production` with **no** `NEXT_PUBLIC_*` → `@web/next:build` now **fails** (`Invalid environment variables`) instead of completing with a dummy. CI-style (throwaway public URLs, no server secrets) → `turbo build check-types` 14/14.

nrjdalal:
## Update — scoped `SKIP_ENV_VALIDATION_SERVER` (89c4410c)

Per review follow-up: declaring throwaway `NEXT_PUBLIC_*` in the CI workflows was counterintuitive. Replaced that with a **server-scoped flag** so the split is expressed once, in code:

- **base `SKIP_ENV_VALIDATION`** → server + client both fall back (CI/Docker, no real env — workflows keep just this flag, no per-var values).
- **`SKIP_ENV_VALIDATION_SERVER`** → server-only fallback; `NEXT_PUBLIC_*` stay required and validated. The web build sets this (`vercel.json`).

Client polyfill is back but gated on the **base** flag only, so it never fires on the web deploy. Reverted the workflow `NEXT_PUBLIC_*` additions. Verified: web build with the server flag but no client vars → `@web/next` build fails (client strict); CI with base flag + clean env → 14/14, no per-var declarations.

nrjdalal:
## Empty-string edge fixed (d6c87954)

Good catch on the `""` case. Took the fix: `polyfillServer`/`polyfillClient` now use `||` instead of `??`, so an empty-string var counts as absent and falls back to the dummy under a skip flag — coherent with `emptyStringAsUndefined: true`, which already coerces `""` → `undefined`. Previously a var set to `""` under SKIP would coerce to undefined and fail instead of falling back.

Runtime (no flag) still fails on a blank required secret — verified: `POSTGRES_URL=""` with no flag → throws; `POSTGRES_URL=""` + `SKIP_ENV_VALIDATION_SERVER` → db builds via dummy. Client-strict and CI paths unregressed (14/14).

nrjdalal:
## Skill drift fixed (5c15ba43)

Correct catch, and per CLAUDE.md it belongs in this PR — fixed it directly rather than leaving it. `.agents/skills/docker-test/SKILL.md` no longer claims `SKIP_ENV_VALIDATION` skips zod defaults/transforms; since `skipValidation` is gone and validation always runs, defaults and transforms still apply. Verified against the built env:

```
SKIP_ENV_VALIDATION=true, no HONO_PORT / HONO_TRUSTED_ORIGINS:
  HONO_PORT = 4000 (number)          # .default(4000) applied
  HONO_TRUSTED_ORIGINS = ["https://polyfill.url"]  # transform applied, isArray: true
```

Also added a line noting the web deploy's `SKIP_ENV_VALIDATION_SERVER` scope. Only `docker-test` referenced the flag (grepped all skills); it's symlinked into `.claude`/`.github`, so the one edit covers every view.

PR #674, opened 2026-07-11
feat(api): serve WebSockets on Vercel via the Node adapter

nrjdalal:
## Why

`api/hono` imports `upgradeWebSocket`/`websocket` from `hono/bun`, which is coupled to `Bun.serve()`. Vercel Functions don't run `Bun.serve()` (even under the Bun runtime), so `/api/health/ws` never upgraded on Vercel and silently fell back to REST polling. Vercel shipped [native WebSocket support](https://vercel.com/docs/functions/websockets) (public beta, 2026-06-22) whose documented Hono path is the Node adapter (`@hono/node-server` + `ws`), exporting the http server so the platform drives the upgrade.

## What

Pick the WS adapter at boot from `process.env.VERCEL`:

- **Bun (local, Docker/self-host)** keeps the existing `Bun.serve()` path via `hono/bun`, byte-for-byte the current behavior, zero regression.
- **Vercel** uses `@hono/node-server`'s `serve({ fetch, websocket: { server: new WebSocketServer({ noServer: true }) } })` and exports the http server, matching Vercel's official Hono example.

The `/api/health/ws` handler body is unchanged. Adds `@hono/node-server`, `ws`, `@types/ws` to the catalog + `api/hono`. Docs (`type-safe-api`, `deployment/vercel`) and the `api-endpoint` skill updated.

## Verification

- `check-types`, `oxlint`, `oxfmt`, both bundles (`build` + `build:vercel`), and the full pre-commit build (incl. strict docs build + `next build`) all pass.
- **Bun.serve() branch** (default): WS streams snapshot + heartbeat. No regression.
- **Vercel branch, forced under Bun** (`VERCEL=1 bun src/index.ts`): `@hono/node-server` + `ws` boots and streams snapshot + heartbeat identically.
- **Browser e2e**: the marketing home `ApiStatus` badge shows the live WS pulse (`animate-pulse` on the dot = frame from the socket, not REST).
- **Vercel preview (confirmed)**: the platform upgrades the WebSocket under the kept **Bun runtime** (`bunVersion` in `api/hono/vercel.json`), streaming the live pulse. No Node-runtime switch needed: the Node adapter runs on Vercel's Bun runtime via its `node:http`/`ws` compat, and `serve()`'s self-`listen()` is tolerated (Vercel's own docs example self-listens too).

api-status live pulse via node adapter

## Notes (previously open, now resolved)

- **Runtime**: kept the **Bun runtime** in `api/hono/vercel.json`; the preview confirmed Vercel drives the upgrade into the exported server under it, so no switch to the Node runtime was required.
- **Port**: the Vercel branch passes `port: Number(process.env.PORT) || env.HONO_PORT` so a forced local run still binds 4000; on Vercel the value is inert (the platform drives the upgrade regardless), matching how Vercel's example lets `serve()` default the port.
- A follow-up doc commit aligns two wording details with Vercel's page: Fluid compute is "on or after" April 23 2025, and a connection closes when the function reaches its max duration rather than a fixed "few minutes".

## Incidental (not part of the feature)

`bun install` reconciled a stale `packages/cli` version in `bun.lock` (`0.0.24` -> `0.0.25`) to match `packages/cli/package.json`, which is already `0.0.25` on `canary`. Pre-existing lockfile drift on `canary`, not introduced here.

### DIFF
--- a/.agents/skills/api-endpoint/SKILL.md
+++ b/.agents/skills/api-endpoint/SKILL.md
@@ -87,12 +87,12 @@ Client components reading REST data use TanStack Query (see `components/common/a
 
 ## WebSocket routes
 
-For a live server-to-client stream instead of polling, upgrade a `GET` with `upgradeWebSocket` from `hono/bun` and add the shared `websocket` handler to the Bun server export next to `fetch` (`api/hono/src/index.ts`). `/api/health/ws` is the reference: a snapshot on connect, then a heartbeat every 5s.
+For a live server-to-client stream instead of polling, upgrade a `GET` with `upgradeWebSocket` (`api/hono/src/index.ts`). The socket owner differs by host: on Bun (local, Docker) it's `hono/bun` with the shared `websocket` handler next to `fetch` in the `Bun.serve()` export; on Vercel it's the Node adapter (`@hono/node-server` + `ws`) exporting the http server, since Vercel Functions can't run `Bun.serve()`. The starter picks the adapter at boot from `process.env.VERCEL`, so a new WS route just registers under the selected `upgradeWebSocket`. `/api/health/ws` is the reference: a snapshot on connect, then a heartbeat every 5s.
 
 - The typed client reaches it with `apiClient.health.ws.$ws()`, a standard `WebSocket` pointed at the API base (`http` becomes `ws`).
 - Frames are not RPC-typed: `ws.send()` takes a raw string and `$ws()` returns a plain `WebSocket`. Parse defensively and read only the fields you need; don't hand-maintain a shared payload type RPC can't derive.
 - Keep a `describeRoute` so the upgrade lists in Scalar as a `101`, and describe the frame shape in the route `description`, since OpenAPI can't schema-type WS frames and there is no `{ data }`/`{ error }` envelope.
 - The handshake skips `cors()` (browsers don't apply CORS to WebSockets) and `$ws()` sends no credentials, so gate a sensitive route on the `Origin` header or a token inside the handler, not the allowlist. `/api/health/ws` serves public data, so it doesn't.
-- `bun --hot` picks up edits to an existing `index.ts` route, but restart the stack if `hono/bun` isn't yet wired into the Bun export.
+- `bun --hot` picks up edits to an existing `index.ts` route, but restart the stack if the `upgradeWebSocket` isn't yet wired into the exported server.
 
-Reference client: `components/marketing/api-status.tsx`. REST `/api/health` is the always-honest baseline, polled whenever no frame is live; the socket overlays a live pulse and reconnects with capped backoff, so a serverless deploy or transient blip degrades to the REST-polled state instead of a broken badge.
+Reference client: `components/marketing/api-status.tsx`. REST `/api/health` is the always-honest baseline, polled whenever no frame is live; the socket overlays a live pulse and reconnects with capped backoff, so a cold start or transient blip (Vercel caps a connection at a few minutes) degrades to the REST-polled state instead of a broken badge.
--- a/.agents/skills/api-endpoint/SKILL.md
+++ b/.agents/skills/api-endpoint/SKILL.md
@@ -87,7 +87,7 @@ Client components reading REST data use TanStack Query (see `components/common/a
 
 ## WebSocket routes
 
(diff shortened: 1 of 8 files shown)


nrjdalal:
Thanks for the deep review, addressed the actionable items:

**Refactored per feedback (don't pollute `index.ts`)** — the host/runtime plumbing (Vercel-vs-Bun adapter selection, `WebSocketServer`/`serve()` wiring, dual server-export shape) moved into a new `api/hono/src/lib/server.ts` exporting `upgradeWebSocket` + `createServer(app)`. `index.ts` now just imports both, registers the route, and ends with `export default createServer(app)`. No behavior change.

**Comment nits fixed:**
- The `onClose` comment no longer claims Bun "has no error event" as if single-host — generalized to both adapters (close fires on every disconnect; `ws` always emits close after error).
- The export comment no longer says Vercel "owns the port" or frames the change as WS-only — now: the Node export governs *all* traffic, binds `PORT` when Vercel sets it, else `HONO_PORT`.
- Left the `as typeof bunUpgradeWebSocket` cast (well-commented); noted your point that it masks divergence if a future WS route uses `onMessage`/`onError`/`ws` beyond `send`.

**Full-surface local check (your must-do, done to the extent possible off-Vercel)** — re-smoked both branches after the refactor. On the `VERCEL=1` node export under Bun, REST is intact alongside WS:

```
/api/health      -> 200
/api/openapi.json -> 200
/api/waitlist    -> 200
+ /api/health/ws  -> snapshot + heartbeat
```

This confirms the export-shape flip doesn't regress REST *locally*. The genuine platform-side unknown (Vercel routing the upgrade + all traffic into the exported listening server) still needs a **preview deploy** to confirm end-to-end — I can't reproduce that off-platform.

Left as-is (you flagged non-blocking): the static both-branches-reachable ternary keeps `@hono/node-server`+`ws` in the self-host bundle. A dynamic `import()` in the `onVercel` branch would shed it at the cost of top-level await; not worth it for a starter.

nrjdalal:
Applied the three actionable nits (`4eef2d65`):

- **`PORT=0` edge** — guard on the raw string, not the coerced number: `process.env.PORT ? Number(process.env.PORT) : env.HONO_PORT`. A legit `PORT=0` (ephemeral bind) now survives instead of silently becoming 4000; empty/unset still falls back. Re-smoked the Vercel branch with no PORT set: binds 4000 + WS streams.
- **Doc snippet imports** — the `lib/server.ts` block in `type-safe-api.mdx` now shows the `env` and `type { Hono }` imports it uses, so it's copy-pasteable.
- **Cast caution** — added a bullet to the `api-endpoint` skill: the node→bun `upgradeWebSocket` cast makes `ws` (WSContext) host-specific fields (e.g. `ws.raw`) type green but runtime-divergent on Vercel; stick to `send`/`close` or branch per host.

Acknowledged non-blocking, left as-is: the self-host bundle statically pulling `@hono/node-server`+`ws` (boot-time branching over dynamic import is the right trade for a starter). The one remaining gate is unchanged: the full-surface preview-deploy check.

nrjdalal:
Third-pass items handled (`6b7a4170`):

- **Host selector tightened** to `process.env.VERCEL === "1"` (the canonical Vercel value) instead of `Boolean(...)`, so a stray `VERCEL="0"`/`"false"` can't take the node branch. Re-smoked both branches: `VERCEL=1` → node adapter + WS, default → `Bun.serve()` + WS.
- **CLI-lockfile note added to the PR body** — the `bun.lock` `packages/cli` `0.0.24 → 0.0.25` is `bun install` reconciling pre-existing `canary` drift (package.json is already `0.0.25`), not a feature change.

Left as-is (your nits, deliberately): `Number(process.env.PORT)` NaN-on-non-numeric is guarded by the truthy check and settled by open-question #2 at preview; the bundle dead-weight is the right boot-time-branching trade for a starter.

Gate unchanged: the full-surface preview-deploy confirmation.

nrjdalal:
## ✅ Validated on real Vercel: the merge gate is cleared

The one thing three review rounds flagged as unverifiable locally, whether Vercel actually routes the WebSocket upgrade into the exported `http.Server`, is now confirmed on a **real Vercel deployment** of this branch (`0.0.98-6b7a417`), temporarily aliased to the canary api domain (since reverted).

**WebSocket upgrade works on Vercel:**
```
wss://<preview>/api/health/ws  ->  OPEN + snapshot + heartbeat ✓
```

**Full REST surface on the same deployment (the export flip governs all traffic):**
```
/api/health       200
/api/openapi.json 200
/api/docs         200
/api/waitlist     200
/api/auth/ok      200
/api/v1/user      401   (correct: unauthenticated)
/api/headers      404   (correct: that route is at root /headers, not /api)
```
No 5xx anywhere, so the switch from `{ fetch }` to a listening `http.Server` regresses no route.

**End-to-end in a browser:** the marketing `ApiStatus` badge showed the live WS pulse (`animate-pulse` = frame from the socket, not REST fallback) when pointed at this deployment.

### Both open questions resolved
1. **Runtime:** it works under the existing **Bun runtime** (`bunVersion 1.3.14`) with the node adapter. No switch to the Node runtime needed.
2. **Port / export shape:** binding the `http.Server` at import did not regress REST or WS.

Nothing else outstanding on my side: review nits addressed, docs + `api-endpoint` skill synced, builds green. Ready to merge into `canary`.

PR #681, opened 2026-07-11
perf(web): strip the shiki grammar bundle to content-used languages

nrjdalal:
## What

The marketing landing page imported `codeToHtml` from `"shiki"`, whose `bundledLanguages` statically references every grammar. Next's output-file-tracing follows that import graph, so **all ~250 grammars (~10 MB)** traced into `.next/standalone` even though highlighting runs only at build time. (Docs/blog code blocks are baked to static HTML by fumadocs at build and ship zero runtime grammars; the landing page was the only runtime highlighter.)

This makes the bundle **content-driven** instead of hardcoded.

- **`.github/scripts/shiki-langs.ts`** (new): scans every code fence under `web/next/content/**`, resolves aliases via shiki's own `bundledLanguagesInfo` (`bash`→shellscript, `ts`→typescript, ...), and generates `web/next/src/lib/shiki-languages.ts` — a lazy grammar map keyed by id **and** every alias. Git-ignored; regenerated in `dev`, `build`, and `check-types`, so a code block in a new language just works after a rerun.
- **`(marketing)/page.tsx`**: builds a custom shiki bundle with `createBundledHighlighter` + `createSingletonShorthands` over the generated map. The `highlight()` call site is unchanged; `codeToHtml`'s `lang` is now type-checked against the content-derived set.
- Adds `@shikijs/langs` + `@shikijs/themes` as catalog deps (already transitive; now honest, needed for the generated `import()` targets to resolve).
- Documents the mechanism in a **`shiki-langs` skill** and the AGENTS.md skill table.

## Why not the alternatives

- **`DynamicCodeBlock`** highlights in the browser and ships shiki + grammars to the *client* bundle (with a Suspense flash). It's for code not known at build; ours is four static literals.
- **`ServerCodeBlock` / `fumadocs-core/highlight`** pull `shiki/full`, re-tracing the whole ~250-grammar bundle.

## Verified

- **standalone 76M → 65M** (`@shikijs/langs` no longer traced; 9 grammars inlined into the marketing chunk).
- No leak: unused grammars (python/rust/...) appear only as embedded-scope strings inside the markdown grammar, never as real grammars.
- Coverage complete: every content fence token maps to a generated key.
- Output **byte-identical** (same grammars, themes, `colorReplacements`): 94 dual-theme `--shiki-light`/`--shiki-dark` token spans, github-dark comment replacement applied 14x.
- Lint clean, `check-types` 12/12, format correct, production build passes.
- Browser: all four hero code blocks highlight correctly. landing


nrjdalal:
Parking this for now, to be revisited later. Converted to draft until then.

nrjdalal:
Parking this for now, we'll revisit stripping the Shiki grammar bundle later. Branch kept (`perf/shiki-content-langs`) so this can be reopened as-is.

PR #690, opened 2026-07-12
feat(cli): preview the --canary no-op in the init dry-run

nrjdalal:
## What

Closes the follow-up flagged in the #682 retrospective review: `init --dry-run` against an existing ZeroStarter checkout printed `mode: in place` with no hint that `--canary` would be ignored (only the real run logs the warning). The dry-run now previews the no-op so the plan matches the action.

```
$ zerostarter init <existing-checkout> --canary --dry-run
bunx zerostarter init (dry run)
  target: …
  name:   …
  mode:   in place
  note:   --canary ignored, converting in place (nothing is fetched)   <- new
```

Only prints when the target is an in-place convert *and* `--canary` was passed; a normal `fetch` plan and an in-place plan without `--canary` are unchanged.

## Tests

`init.test.ts` gains a case that scaffolds an in-place checkout (a `packages/config/src/site.ts`) and asserts the dry-run shows both `in place` and `--canary ignored`. The helper now takes an optional `setup` callback (explicit `if` guard, not optional chaining, per the repo convention).

- `bun test` (packages/cli): **93 pass, 0 fail** · `check-types` clean.
- Verified the real `--dry-run` output (with and without `--canary`).

## Note

This is a genuine `packages/cli/**` change, so on merge to `canary` it also exercises the **corrected canary prerelease version** from #688 - it should publish `0.1.2-canary.0` (patch-based), not the old `0.2.0-canary.0`.

### DIFF
--- a/packages/cli/bin/commands/init.ts
+++ b/packages/cli/bin/commands/init.ts
@@ -122,11 +122,17 @@ export const init = async (argv: string[]) => {
   const name = basename(target)
   const brand = { name }
 
+  const canaryInPlaceNote =
+    "--canary ignored: converting the existing checkout in place, so nothing is fetched."
+
   if (values["dry-run"]) {
     console.log("bunx zerostarter init (dry run)")
     console.log(`  target: ${target}`)
     console.log(`  name:   ${name}`)
     console.log(`  mode:   ${isZerostarter(target) ? "in place" : `fetch ${ref}`}`)
+    if (isZerostarter(target) && values.canary) {
+      console.log(`  note:   ${canaryInPlaceNote}`)
+    }
     return
   }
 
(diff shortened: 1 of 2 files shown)


nrjdalal:
Addressed all three review suggestions:

1. **Wording parity** - both the dry-run note and the real-run `logWarn` now come from a single `canaryInPlaceNote` const, so the preview reads verbatim like the log it foreshadows (`--canary ignored: converting the existing checkout in place, so nothing is fetched.`).
2. **Negative test** - added a case asserting the note is *absent* for an in-place plan without `--canary`, locking the "only prints when `--canary`" contract.
3. **Independent assertions** - both in-place tests now assert `mode:   in place` explicitly (not the bare `in place` the note also contains), so the mode line and the note are verified separately. Deduped the scaffold into a `scaffoldCheckout` helper.

`bun test` 94 pass, `check-types` clean, dry-run output re-verified.

PR #695, opened 2026-07-13
refactor(api): gate the agent sign-in route behind AGENT_AUTH_SECRET

nrjdalal:
Tighten the local-only agent sign-in so it is opted into deliberately, one of the [hardening refactors](https://github.com/nrjdalal/zerostarter/blob/canary/.github/notes/plans/hardening-refactors.md) from the external [evaluation](https://github.com/nrjdalal/saas-starter-evals).

## Background
`/api/agents/sign-in-as` mints an admin session for local AI-agent testing (a real DX differentiator, kept). It was gated on `isLocal(NODE_ENV)` alone, and `.env.example` ships `NODE_ENV=local`, so a default self-host could keep the route reachable. Tighten the gate.

## Change
- Add `AGENT_AUTH_SECRET` (optional, no polyfill) to `packages/env` + `.env.example` (empty).
- Mount the router only when `isLocal(NODE_ENV) && env.AGENT_AUTH_SECRET`. The Origin check still guards each request.
- The secret is **unset by default** and the CLI seed doesn't fill it, so a fresh clone and any deployed default env mount no admin route. A dev opts into agent login by setting it once (`openssl rand -base64 32`).
- Docs synced: `AGENTS.md`, the `dev` skill, and `working-with-agents` / `setup` / `authentication`.

## Verified (runtime, both ways)
| | secret unset | secret set |
| --- | --- | --- |
| `POST /api/agents/sign-in-as` (`NODE_ENV=local`) | **404** | **302** + `role: admin` session |
| untrusted Origin (secret set) | - | rejected |

The automated route test is deferred to the product-test harness (`bun test` auto-loads `.env`, so it would false-pass locally); the gate is runtime-verified above.

## Summary by CodeRabbit

* **New Features**
  * Added a dev-only **agent sign-in** capability gated by `AGENT_AUTH_SECRET` in local `.env`.
  * **Login (agents)** is now driven by server-provided sign-in options and only appears when the agent method is enabled.
* **Bug Fixes**
  * Sign-in option availability is now consistent with the enabled authentication methods returned by the server.
* **Documentation**
  * Updated setup and agent workflow docs to include `AGENT_AUTH_SECRET` preconditions and the local-only gating behavior (returns 404 when not enabled).
  * Updated auth documentation to reflect how providers are discovered.
* **Tests**
  * Added coverage for `AGENT_AUTH_SECRET` seeding and non-overwrite behavior.
* **Chores**
  * Updated environment seeding guidance for local scaffolds.

PR #698, opened 2026-07-13
feat: gate the local agent sign-in behind an AGENT_SIGNIN_ENABLED toggle

nrjdalal:
## Summary

Rebuilds the agent sign-in hardening from scratch on `canary`, **superseding #695**. The local-only agent sign-in (`POST /api/agents/sign-in-as`, which mints a `LocalAgent` admin session) was gated on `NODE_ENV=local` alone, and since `.env.example` ships `NODE_ENV=local`, a default self-host could keep the admin-minting route reachable. This adds an explicit **`AGENT_SIGNIN_ENABLED`** toggle (a real boolean, off by default) as a second gate, and keeps the out-of-the-box agent DX by seeding it on `init`.

Named `AGENT_SIGNIN_ENABLED`, not `AGENT_AUTH_*`: it is a boolean toggle for one local dev shortcut, not an authentication mechanism and not a rotatable secret.

## What changed

- **env** (`packages/env`): `AGENT_SIGNIN_ENABLED: z.stringbool().default(false)`; blank / `false` / unset -> off. `.env.example` ships it blank.
- **api** (`api/hono`): one shared `agentSignInEnabled()` predicate (`isLocal(NODE_ENV) && env.AGENT_SIGNIN_ENABLED`) gates both the route mount (`agents.ts`) and the `/providers` advertisement (`auth.ts`), so the button can never drift from the route it posts to.
- **web**: the sign-in dialog renders **Login (agents)** only when the API advertises the `agent` provider, behind an extra `isDev` guard so a production bundle never renders a dev-only admin control; the empty state gains a dev-only hint.
- **cli**: `seedEnv` writes `AGENT_SIGNIN_ENABLED=true` into the scaffolded (gitignored) `.env`, so `bunx zerostarter init` yields a working button out of the box; it only fills empty keys and never touches the committed `.env.example`.
- **docs**: setup, working-with-agents, authentication, environment, `AGENTS.md`, the `dev` and `ui-verify` skills, and the hardening plan note.

## Security posture

The toggle ships **off** (blank) in the committed `.env.example`, so a `git clone` or any deploy exposes no admin-minting route. `init` writes `=true` only into the local gitignored `.env`. The `isLocal(NODE_ENV)` gate is an independent second lock, so a seeded toggle is inert on any real deploy. Exposure now requires two compound misconfigurations instead of one.

## Verification

- `bun run check-types` 12/12 (incl. `@web/next` RPC inference and `@api/hono`); `bun run lint` clean; CLI `bun test` 9/9.
- `z.stringbool().default(false)`: unset/blank -> `false`, `"false"` -> `false` (no truthy-string footgun), `"true"` -> `true`.
- `seedEnv` against the real `.env.example`: writes `AGENT_SIGNIN_ENABLED=true` on one line, in place.
- **Runtime, both directions** (booted the API):
  - `=true`: `/api/auth/providers` -> `{"providers":["agent"]}`; `POST /sign-in-as` -> `500` (route **mounted**).
  - `=false`: `/api/auth/providers` -> `{"providers":[]}`; `POST /sign-in-as` -> `404` (route **unmounted**).
- The **Login (agents)** button renders from that runtime-verified `/providers`; the client mapping is unchanged from #695 (browser-verified). A full browser screenshot pass can be run before merge if wanted.

## Supersedes

Replaces #695 (which used a presence-checked `AGENT_AUTH_SECRET`). Recommend closing #695 in favor of this.

## Summary by CodeRabbit

* **New Features**
  * Added an optional local-development setting to enable agent sign-in.
  * Local setup now enables agent sign-in automatically for development.
  * Sign-in options now display the agent login method only when enabled.
  * The “Login (agents)” option is hidden when unavailable, with guidance provided for enabling it.

* **Documentation**
  * Updated setup, authentication, environment, and agent workflow documentation with the new requirements and safeguards.

* **Bug Fixes**
  * Prevented agent sign-in controls from appearing when the sign-in route is unavailable.

### DIFF
--- a/.agents/skills/dev/SKILL.md
+++ b/.agents/skills/dev/SKILL.md
@@ -39,11 +39,12 @@ Restart the same way after changing `@packages/*` exports the API consumes; they
 
 ## Agent login
 
-Sign in as `LocalAgent` (local only, trusted Origin required):
+Sign in as `LocalAgent` (local only, trusted Origin required). The route is gated on `AGENT_SIGNIN_ENABLED`: set it to `true` in `.env` first, or the route 404s. It is off by default, so a fresh clone and any deploy expose no admin-minting route.
 
 ```bash
+grep -q '^AGENT_SIGNIN_ENABLED=true' .env || echo "AGENT_SIGNIN_ENABLED=true" >> .env   # once, to enable
 curl -sS -c cookies.txt -X POST -H "Origin: http://localhost:3000" http://localhost:4000/api/agents/sign-in-as
 curl -sS -b cookies.txt http://localhost:4000/api/v1/user
 ```
 
-In the browser: click **Login** in the top navbar (hidden on `/console` and `/dashboard`), then **Login (agents)** in the dialog (development only).
+In the browser: click **Login** in the top navbar (hidden on `/console` and `/dashboard`), then **Login (agents)** in the dialog (development only, with `AGENT_SIGNIN_ENABLED=true`).
--- a/.agents/skills/dev/SKILL.md
+++ b/.agents/skills/dev/SKILL.md
@@ -42,7 +42,6 @@ Restart the same way after changing `@packages/*` exports the API consumes; they
 Sign in as `LocalAgent` (local only, trusted Origin required). The route is gated on `AGENT_SIGNIN_ENABLED`: set it to `true` in `.env` first, or the route 404s. It is off by default, so a fresh clone and any deploy expose no admin-minting route.
 
 ```bash
-grep -q '^AGENT_SIGNIN_ENABLED=true' .env || echo "AGENT_SIGNIN_ENABLED=true" >> .env   # once, to enable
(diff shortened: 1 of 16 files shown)


nrjdalal:
## Browser verification (agent-browser, live dev stack)

Drove the real sign-in dialog at `localhost:3000` against a running API, in both toggle states (viewport 1782x972). This closes the one item both reviews flagged.

**`AGENT_SIGNIN_ENABLED=true`** — `/api/auth/providers` advertises `["agent"]`, and the dialog renders the **Login (agents)** button:

sign-in dialog with toggle on

**`AGENT_SIGNIN_ENABLED=false`** — `/api/auth/providers` returns `[]`, no button, and the dev-only empty-state hint renders verbatim (*"No sign-in options are configured yet. Set AGENT_SIGNIN_ENABLED=true in .env to enable local agent login."*):

sign-in dialog with toggle off

Matches the earlier API-level runtime checks (route mounts / 404s on the same toggle).

PR #700, opened 2026-07-13
refactor(api): generate OpenAPI schemas from the Zod validators

### LINKED ISSUE #664: OpenAPI: WS upgrade route /api/health/ws lists inapplicable 429/500 responses in Scalar
## Summary

The WebSocket upgrade route `/api/health/ws` (added in #661) is documented in the Scalar reference (`/api/docs`) with `application/json` `429`/`500` error-envelope responses, alongside its `101` Switching Protocols. Those JSON error bodies don't apply to a WebSocket handshake, so the listing is cosmetically inaccurate.

## Root cause

`openAPIRouteHandler` in `api/hono/src/index.ts` sets:

```ts
defaultOptions: {
  GET: { responses: globalErrorResponses },
  POST: { responses: globalErrorResponses },
}
```

These global error responses are injected onto **every** GET operation, including the WS upgrade route, whose `describeRoute` only declares `101`. hono-openapi merges the defaults in, so the route shows `101` + `429` + `500`.

## Impact

Cosmetic / doc-accuracy only. No functional issue:
- The `429` is in fact reachable (the rate limiter runs on the upgrade GET request).
- The `500` is the shared global default.
- The mainline behavior and the `$ws()` client are unaffected.

It just looks odd to see JSON error bodies next to a `101` upgrade in Scalar.

## Options

1. Leave it (documenting a WebSocket inside an HTTP spec has this seam inherently; the endpoint stays discoverable with its `$ws()` code sample).
2. Add a per-route mechanism to opt `/health/ws` out of the shared `defaultOptions.GET` responses (no clean hook for this in the current wiring).
3. Stop using `defaultOptions.GET`/`POST` and spread `globalErrorResponses` explicitly per route (verbose; contradicts the reason that default exists).
4. Drop `/health/ws` from the OpenAPI spec entirely (loses the Scalar discoverability the endpoint was given on request).

## Context

Deferred from #661 as a non-blocking cosmetic note, surfaced across multiple review passes there. See the `Live data over WebSockets` section in `web/next/content/docs/getting-started/type-safe-api.mdx` and the WebSocket routes section in `.agents/skills/api-endpoint/SKILL.md`.

nrjdalal:
## What

Implements **P2-2** from the [saas-starter-evals](https://github.com/nrjdalal/saas-starter-evals) audit: *"~38% of benchmark router LOC was hand-written `describeRoute` OpenAPI duplication."* Every documented route hand-wrote the same ~30-line scaffold (the `z.object({ data })` wrapper, the `200/OK` content block, the `x-codeSamples`, the error-response spreads) next to its Zod validator, a drift risk the machine can produce.

Two helpers in `api/hono/src/lib/route.ts` now generate it:

- **`jsonRoute(spec)`** builds the `describeRoute` doc: wraps `output` in the `{ data }` envelope for the 200, expands the `hono/client` sample from one call expression, and attaches only the error responses that apply (429/500 always; `+400` when `validated`, `+401` when `auth`).
- **`jsonBody(schema, message)`** swaps `@hono/standard-validator`'s `sValidator` for hono-openapi's own `validator`, so the OpenAPI `requestBody` is generated from **the same schema it validates at runtime** and the two cannot drift.

`/health`, `waitlist`, and `v1` are migrated; the now-unused `@hono/standard-validator` dependency is dropped. Since `jsonRoute` owns each route's error responses, the global `defaultOptions` 429/500 injection is removed.

`jsonRoute` + `jsonBody` (not a single `defineRoute(method, path, io, handler)`) keeps the native `.get/.post(path, ...)` chain, which preserves Hono RPC type inference a monolithic wrapper would fight. `AppType` and the web client are unchanged.

## Impact

- **31 insertions vs 149 deletions** across the routers + `index.ts` (a ~55-line reusable helper added).
- Two spec improvements fall out, confirmed by a byte-diff of `/api/openapi.json` (baseline canary vs this branch) showing **exactly** these two deltas and nothing else:
  1. `/api/health/ws` now lists only its `101` instead of inapplicable `429`/`500` (**resolves #664**).
  2. `/api/waitlist` POST documents its `requestBody` (`email` required + `subject`) for the first time, generated from the validator.

### Before / After (Scalar `/api/docs`)

| | Before (canary) | After |
|---|---|---|
| WS route responses | before-ws `101 429 500` | after-ws `101` only |
| Waitlist POST | before-post no request body | after-post `Body: email, subject` |

## Verification

- `check-types`, `oxlint`, `oxfmt --check`, strict docs gate, **109 tests** — all green; full pre-commit build passed.
- Runtime curls: `GET /health`, `GET/POST /waitlist` (valid `{ data }` + invalid `VALIDATION_ERROR`), `GET /v1/user` unauth `401`.
- Browser (agent-browser): Scalar renders the new spec; homepage api-status pill green ("All systems are operational"), so the health WebSocket still streams after the `index.ts` change.

## Docs synced

`api-endpoint` skill (new canonical pattern), `type-safe-api.mdx`, `api-conventions.mdx`, and the plan notes (`api-envelope-typed-endpoint.md` move 2 marked shipped; `openapi-ws-responses.md` removed as resolved).

## Follow-ups (not in this PR)

Move 1 of the parent plan (a shared `Envelope<T>` type across `error.ts` and `client.ts`) and optionally tightening the response side with hono-openapi's `describeResponse`.

## Summary by CodeRabbit

- **New Features**
  - Added standardized JSON route helpers for typed responses, request validation, and consistent error envelopes.
  - Improved API reference generation with accurate response schemas and route-specific error responses.
  - Added typed client code samples to API documentation.

- **Documentation**
  - Updated API implementation and validation guidance to reflect the new route helpers.
  - Corrected documentation for WebSocket and API reference behavior.


nrjdalal:
## Regression investigation: `sValidator` → hono-openapi `validator`

Flagged that the API deliberately moved off hono-openapi's `validator` to `@hono/standard-validator`'s `sValidator` early on (PR #287 dropped the validator-driven query example; #523 landed real validation on `sValidator`) because an older `hono-openapi` **regressed Hono RPC inference** — the typed client, the eval's "crown jewel." This PR re-adopts that validator, so before keeping it I re-verified zero regression on the current pin (`hono-openapi@1.3.1`), running both implementations side by side:

| Dimension | Method | Result |
|---|---|---|
| **RPC types (crown jewel)** | full byte-diff of `api/hono/dist/index.d.mts` (`AppType`), sValidator vs validator | **byte-identical** (`diff` exit 0) |
| RPC input + output | `@ts-expect-error` negative tests: `$post({})`, `email: 123`, `data.message: number` | all correctly rejected; correct shape compiles |
| **Runtime validation** | 12-case battery vs canary: valid, bad-format, missing, empty, whitespace-`.trim()`, `>254` cap, extra field, wrong type, malformed JSON, form-encoded, no content-type, GET count | **byte-identical** responses |
| OpenAPI spec | byte-diff vs canary | only the 2 intended deltas (requestBody added, WS 429/500 removed) |
| Bundle | `bundle/index.mjs` size | **641 bytes smaller** |

Conclusion: the historical degradation **does not reproduce** on the current pin, so we keep the lower-boilerplate `validator`. The rationale + a guard-rail (re-diff `dist/index.d.mts` on any `hono-openapi` bump; revert to `sValidator` if the typed client changes) is now recorded in the `api-endpoint` skill and the plan note (commit `930495fd`).

nrjdalal:
Triage against current `canary` (2026-08-05): **partly landed, scope needs re-checking before rebase.**

`resolver()` from `hono-openapi` is already in use across the routers on `canary` (`admin.ts`, `v1.ts`, `waitlist.ts`), and `/api/openapi.json` serves a valid OpenAPI 3.1.0 document with 11 documented paths (verified live this session against the Docker image). So the "derive schemas from the Zod validators" idea is at least partially in place.

What is worth confirming before rebasing: which endpoints still hand-write their response shapes rather than deriving them, since that is the remaining value here. The related backlog entry is `.github/notes/plans/api-envelope-typed-endpoint.md`.

Status: `CONFLICTING`, untouched since 2026-07-25.

Issue #719, opened 2026-07-14
vercel.app deploys: public-suffix cookies break auth; document the same-origin INTERNAL_API_URL topology and its WebSocket escape hatch


nrjdalal:
## Summary

Deploying the template as two Vercel projects **without a custom domain** (plain `*.vercel.app` URLs) breaks sign-in out of the box, and the documented env matrix in `docs/deployment/vercel.mdx` points at the broken topology. The root cause is that `vercel.app` is on the Public Suffix List, so no cookie can ever be shared between the web and api projects. The working setup routes all API traffic through the web origin via the existing `/api/:path*` rewrite, with `INTERNAL_API_URL` breaking the resulting self-loop. That topology then breaks the WebSocket path (rewrites cannot proxy upgrades), which needs its own escape hatch. All of this was hit and solved in the field while deploying ZeroStudio (a scaffold of this template); details and fixes below so the template can absorb them.

## Field report, in the order the failures appeared

1. Followed the current `vercel.mdx` matrix: two projects, `HONO_APP_URL=https://<api>.vercel.app`, GitHub OAuth callback on the api origin. Sign-in redirected back with `?error=state_mismatch`.
2. Cause: `Set-Cookie` carried `Domain=.vercel.app`. `getCookieDomain()` strips one label and returns the parent domain, but `vercel.app` is a **public suffix**, so browsers silently reject the cookie. The OAuth `state` cookie never persists, hence `state_mismatch`.
3. Patched `getCookieDomain` to return `undefined` (host-only cookies) on public hosting suffixes. Sign-in on the api origin then works, but the **web** origin still cannot see the session: there is no shared parent domain to ride on, so the cross-origin topology is unfixable on `vercel.app` by design.
4. Switched to the same-origin topology the template already half-supports: browser only ever talks to the web origin, and the existing rewrite (`/api/:path*` -> `INTERNAL_API_URL || NEXT_PUBLIC_API_URL`) proxies to the api project server-side.
5. First attempt 508-looped: with `NEXT_PUBLIC_API_URL` set to the web origin (required, so the client and auth are same-origin), the rewrite destination defaults to the web origin too, i.e. itself. `INTERNAL_API_URL=https://<api>.vercel.app` breaks the loop. Nothing in the docs or a runtime guard hints at this; the failure shows up as `508 Loop Detected` on every `/api/*` call (the landing page health badge showing "Systems are facing issues" was the only visible symptom).
[...]

Issue #707, opened 2026-07-13
Icebox


nrjdalal:
## What the Icebox is

A holding area for things that have been **raised but not yet triaged into a decision**. Something lands here the moment it is noticed: a review comment, a follow-up carved out of a PR, an idea, a maybe-bug. The point is to keep the context before anyone has decided what to actually do about it.

Being in the Icebox is **deliberately non-committal, in both directions.** An entry here is not a promise that it will be built, and not a ruling that it never will be. It is explicitly *undecided.* That is the whole value: a place to park something honestly without pretending to know its fate yet.

## What an entry might turn out to be

Any of these, and we do not claim to know which up front:

- **Real and worth doing:** it graduates to the plans backlog (`.github/notes/plans/`) and gets scheduled.
- **Real and required:** it turns out to be load-bearing and jumps the queue.
- **Real but not worth it:** a genuine issue we consciously choose to leave (a won't-fix).
- **Not actually a problem:** on a closer look the concern dissolves.
- **A false claim:** the thing it describes is not true of the code at all.
- **Something else:** reframed, merged into another item, or superseded by a later change.

Forcing every raised concern immediately into "planned" or "closed" is the failure this avoids: it either clutters the real backlog with unvetted items, or throws away context we might want later. The Icebox is that missing middle state.

## How it works

- This issue is the at-a-glance **index.** Each checkbox links to a fuller write-up under `.github/notes/plans/`, where the context and the open question live.
- It **stays open permanently.** Add an entry when something goes on ice; check it off when it thaws, and note what happened (shipped, dismissed, or moved to the backlog).
- Entries come from any PR, review, audit, or eval, not only the ones below.
- The order is loose and ever-changing. It is **not** a priority queue, because nothing here is scheduled.

## On ice

- [x] Move the split-deploy decision to build time (stop shipping the suffix predicate) - **GRADUATED to the backlog:** decided - tldts at build (dev-only), env owns the value / auth owns the behavior, wide "no shared cookie domain" trigger; spec written as #721 (ready-for-agent), scheduled on `feat/build-time-deploy-mode`, superseding the runtime-detection half of #720. The "tldts vs curated PSL set" question is resolved (tldts) - `plans/build-time-deploy-mode.md`
- [x] Encode the handoff route behaviors (mode-gate 404, claim single-use/replay, wrong-nonce reject) as CI tests - needs an api integration-test harness - `plans/handoff-route-tests.md`
[...]

Issue #677, opened 2026-07-11
Dynamic per-branch preview URLs on custom domains (wildcard prefix scheme)


nrjdalal:
## Goal

Replace the opaque generated preview URLs (e.g. `zerostarter-op988nll0-nrjdalal.vercel.app`) with predictable, dynamic per-branch URLs on our own domain, for both the web (`zerostarter.dev`) and api (`api.zerostarter.dev`) projects.

Context: while validating #674 on Vercel, a branch was aliased by hand to `canary.zerostarter.dev` / `api.canary.zerostarter.dev`. This issue is to make that automatic and consistent for any branch.

## Naming convention (decided)

Prepend the branch/env as the **leftmost** label of each project's production host:

- web: `<branch>.zerostarter.dev`  (e.g. `canary.zerostarter.dev`)
- api: `<branch>.api.zerostarter.dev`  (e.g. `canary.api.zerostarter.dev`)

Rationale: DNS/Vercel wildcards only match the leftmost label. `*.api.zerostarter.dev` matches `<branch>.api.zerostarter.dev` for any branch, but nothing can dynamically match `api.<branch>.zerostarter.dev` (a middle-label wildcard is unsupported). So the one-off `api.canary.zerostarter.dev` used during #674 should migrate to `canary.api.zerostarter.dev`.

No collision: `*.zerostarter.dev` matches only single-label subdomains, so it does not catch `<branch>.api.zerostarter.dev`, and the explicit `api.zerostarter.dev` production domain wins over the web wildcard by specificity.

### A. Preview Deployment Suffix (native)
- Vercel add-on that swaps the `.vercel.app` suffix for our domain on every preview: `zerostarter-dev-git-<branch>-nrjdalal.preview.zerostarter.dev`.
- Pros: zero maintenance, fully automatic.
- Cons: Pro plan add-on (paid) or Enterprise; **team-wide single suffix** (web + api share it, differentiated by project-slug prefix); keeps the long generated format rather than a clean `<branch>.<host>`; 63-char DNS label limit.

### B. Wildcard domains + alias-on-deploy automation (custom, gives the clean scheme above)
- Add wildcard domains `*.zerostarter.dev` (web) and `*.api.zerostarter.dev` (api).
- Add a CI step (GitHub Action / deploy hook) that runs `vercel alias set <deployment-url> <branch>.<host>` after each preview deploy, deriving `<branch>` from the git ref (slugified).
[...]

PR #720, opened 2026-07-16
feat(auth): deploy-mode resolution with split cookies and a nonce-bound session handoff

nrjdalal:
Part of #719.

Cookie auth breaks when the web and api apps deploy as two projects on a public hosting suffix (plain `vercel.app`, `netlify.app`, and friends): the OAuth state cookie is rejected at storage, first as a `Domain=.vercel.app` cookie (public suffix) and then, host-only, as a `SameSite=Lax` cookie arriving on a cross-site response. Sign-in dies with `state_mismatch` on the template's first-touch deployment path.

## What this does

Resolves a `DeployMode` once at module init (`shared-domain | host-only | split`, pure string work against a curated suffix set, deliberately not a PSL library) and branches everything on that value:

- **shared-domain** (custom domains, portless localhost): the existing `crossSubDomainCookies` path, byte-identical config, byte-identical sign-in redirect chain, `Lax` untouched.
- **host-only** (bare hosts, unknown shapes): today's fallback, unchanged.
- **split** (two projects on a public suffix): host-only `SameSite=None` cookies, plus a nonce-bound single-use session handoff so server-rendered pages on the web origin can read the session without third-party cookies: after OAuth, the api parks the session cookie value under an opaque one-time id via Better Auth's own verification adapter (a 60s single-use token consumed atomically, the same primitive its one-time-token plugin uses), the nonce folded into the row identifier so a wrong nonce claims nothing, and the web route claims it server-to-server and sets the cookie first-party. Outside split mode both handoff routes, the api router and the web claim route, answer 404 as if absent, and the web's sign-in callback stays the literal `/dashboard` string it is today, decided at module scope.

Each sign-in in split mode mints a nonce held as a first-party web cookie and threaded through the handoff, so only the initiating browser can claim it (the same binding OAuth `state` performs, against login CSRF). The id and nonce are validated against the exact 64-hex shape the web mints, so a malformed value is rejected before any lookup.

Split mode also skips Better Auth's OAuth `state`-**cookie** check. That cookie is written by the api on a cross-site fetch, which Safari blocks and Firefox partitions, so without this sign-in would still die at `state_mismatch` on both browsers before the handoff even runs. With a server session store the `state` is also persisted in the DB (single-use CSPRNG, deleted on parse), which stays the CSRF binding, exactly what Better Auth's own `oauth-proxy` plugin relies on. This keeps split-mode sign-in working across browsers, not just Chrome.

## Zero regressions, by construction

- Operator configuration is one shape everywhere: same OAuth callback origin, same `NEXT_PUBLIC_API_URL`, no new required env vars, no secrets on the web app, on free-tier suffixes and enterprise domains alike.
- No performance change for domain deployments: the mode is a static value; the per-request path gains no instructions in any mode. Split-mode cost is one extra redirect and one insert plus delete, per sign-in only.
- Unit suite covers the `getCookieDomain` table (including every suffix), the full `resolveDeployMode` truth table, and the shared deploy primitives (`isSplitPair`, `isPublicHostingSuffix`, `mintHandoffToken`, and the handoff-token shape): 14 tests.

## Field verification

The exact diff runs on a live two-project test pair (`studiozero-web` + `studiozero-api`), where the full chain was verified: GitHub OAuth completing, handoff claim single-use (replay tested dead), server-gated `/dashboard` returning 200 with the handed-off cookie and 307 without, and a real sign-in confirmed end to end. Wire captures and the complete design trail (five alternatives compared, adversarial review) live on #719. The `skipStateCookieCheck` path that unblocks Safari/Firefox is now **confirmed end-to-end on the live pair**: a real GitHub sign-in in **Safari** (Private window, Prevent Cross-Site Tracking on) lands signed-in on the server-gated `/dashboard`, which is unreachable without the full handoff setting a first-party web-origin cookie. Firefox, less strict on third-party cookies, follows.

## Not in this PR

- `vercel.mdx` full "without a custom domain" walkthrough (follow-up, per #719); this PR ships the split-mode boot warning plus a note in `vercel.mdx` that the two-project `*.vercel.app` setup works out of the box
- Vercel preview self-wiring so a split preview pair auto-configures from the branch-URL pattern (deferred to the #677 plan; branch previews still point at explicit envs, so a feature-branch preview on `*.vercel.app` reuses the production api and CORS-rejects, as expected)
- Per-client rate-limit keying for server-to-server calls (the api's `findIp` keys under the web function's egress IP; forwarding `x-forwarded-for` cannot fix it and is unsafe, so `getSession` stays cookie-only, tracked in the Icebox #707)
- `INTERNAL_API_URL` to `DOCKER_INTERNAL_API_URL` rename (orthogonal, its own commit)
- A2 web-origin session proxy for Safari/Firefox *client-side* session reads: sign-in itself now completes cross-browser (see the `state`-cookie note above); this remaining follow-up is only for browser JS reading the session cross-origin, and its trigger is third-party cookie removal
- Accepted split-mode tradeoff: `skipStateCookieCheck` validates the OAuth callback against the DB state (single-use, server-issued) but not browser-bound, so a relayed callback can set the api-origin session in another browser (a bounded login-CSRF, mostly Chrome third-party cookies; SSR stays protected by the nonce). Inherent to cookie-less cross-origin OAuth, the posture Better Auth's `oauth-proxy` takes; documented at the call site. The tighter callback-to-nonce binding that would close it is a deferred optional hardening (#707)
- The #719 WebSocket escape hatch (`NEXT_PUBLIC_API_WS_URL` + give-up-after-N reconnect backoff to degrade to REST) is orthogonal to this PR's cookie topology and deferred (#707)
- Re-extending the split-mode web session cookie as the session rolls, so SSR does not expire it at the original lifetime on long-lived sessions (iceboxed follow-up, #707)

## Summary by CodeRabbit

* **New Features**
  * Added secure cross-origin session handoff for split web/API deployments using a nonce-based flow, wired through magic-link and social sign-ins.
  * Implemented split-mode routing and session-cookie behavior, including deploy-mode detection (shared-domain vs host-only).
* **Bug Fixes**
  * Refined cookie handling for public hosting suffixes and tightened `getSession` requests to forward only necessary headers.
* **Documentation**
  * Clarified cross-subdomain vs split-deployment sign-in requirements.


nrjdalal:
Superseded by #727 (dissolve / same-origin proxy), merged into canary as 4f4922f4. The nonce-handoff approach here is replaced by routing sign-in same-origin through the web proxy. Closing.

PR #715, opened 2026-07-14
feat(dev): serve local dev through portless named URLs

nrjdalal:
Named `.localhost` dev URLs via [portless](https://portless.sh) (Vercel Labs). `bun run dev` serves the stack on stable named hosts off one **unprivileged** `:1355` HTTP proxy (no sudo, no `/etc/hosts` write), branch-prefixed in worktrees so parallel checkouts never collide on a port.

## What

- `bun run dev` runs the stack through portless: web `http://zerostarter.localhost:1355`, api `http://api.zerostarter.localhost:1355` (branch-prefixed in a worktree, e.g. `feat.zerostarter.localhost`).
- Uses portless the **official/native way** (config mode, `"dev": "portless"` + `portless: {name, script}` with **no `appPort`**), so portless auto-assigns free ports and branch-prefixes worktrees itself.
- The root `dev` script sets `PORTLESS_*` inline and runs turbo; one script `.github/scripts/portless.ts` derives each app's public URLs from `PORTLESS_URL`. `PORTLESS=0 bun run dev` bypasses portless (fixed `:3000`/`:4000`).
- Vendors the official portless skill at `.agents/skills/portless/SKILL.md`.

## Scope (32 files)

Mostly dev tooling + docs, **plus two small, load-bearing runtime fixes the e2e surfaced** (both prod-safe):

- **`packages/auth/src/lib/utils.ts`** — `getCookieDomain`/`getCookiePrefix` gain a `.localhost`-only branch so the web and api subdomains share one cookie behind portless; without it every SSR-gated route (`/dashboard`, `/console`) redirects to home. **Production (`.dev`/`.com`) is byte-for-byte unchanged** (the `.localhost` guard can't fire in prod); a unit test in `packages/auth/test/utils.test.ts` locks down both prod and dev cases.
- **`api/hono/src/lib/server.ts`** — the local Bun.serve branch now honors `process.env.PORT` (portless assigns it), falling back to `HONO_PORT`. Also the standard behavior for self-host platforms that inject `PORT`.
- **`packages/cli/src/convert.ts`** — fork init now rebrands `portless.name` in both app workspaces (web → `<slug>`, api → `api.<slug>`), so a generated fork serves `<slug>.localhost` dev URLs instead of `zerostarter.localhost` (+ test).
[...]

PR #722, opened 2026-07-17
feat(deploy): bake PSL cookie facts at build, decide shared/split/host-only at boot

### LINKED ISSUE #721: Build-time deploy-mode detection (split-deploy)
# Requirements: build-time deploy-mode detection

Status: implemented on `feat/build-time-deploy-mode` (#721; PR pending). Branch: `feat/build-time-deploy-mode` (off `canary`).
Supersedes the detection half of PR #720; is a **rewrite**, not a diff on top of it.
Context and design trail: #719 (umbrella), #720 (the runtime-detection version), Icebox #707.

This document is the single source of truth for two independent spec/implementation passes.
It states _what_ must be true and _which decisions are already locked_; it deliberately leaves
the last mile of mechanics to the implementer except where a decision is called out as locked.

---

## Problem statement

The template ships two deployable apps (`web/next`, `api/hono`). An operator can deploy them as
two separate projects on a shared public hosting suffix (two `*.vercel.app` projects, `*.netlify.app`,
`*.pages.dev`, and friends), or as two different custom domains, or as subdomains of one domain, or
as one shared domain. Cookie-based auth only works out of the box in the shared-domain shapes. In the
others the OAuth state cookie is rejected at storage (a `Domain=.vercel.app` cookie is a public-suffix
cookie the browser drops; host-only it arrives `SameSite=Lax` on a cross-site response), and sign-in
dies with `state_mismatch` on the operator's very first deploy.

PR #720 solved this, and its solution is **field-verified end to end on a live two-project
`*.vercel.app` pair (`studiozero-web` / `studiozero-api`), including a real GitHub sign-in in Safari
with Prevent Cross-Site Tracking on**. But #720 detects which shape it is in **at runtime**, by string-
matching the request/config host against a **hand-curated list of public hosting suffixes**
(`PUBLIC_HOSTING_SUFFIXES`), and it parks those detection primitives in `packages/config`, which is the
wrong home (config is build-tooling + brand identity, not runtime auth behavior). Two problems follow:

1. **The curated suffix list is a maintenance liability and is wrong the moment a new suffix appears.**
   The correct source of truth is the Public Suffix List (PSL). A real PSL library (`tldts`) answers
   "is this host on a public suffix / what is its registrable domain" correctly and forever, but it is
   too heavy to want on the hot path of every request, and nobody wants a PSL dependency shipping in the
   client bundle.
2. **The detection primitives live in the wrong package**, and the mode is recomputed at runtime when
   it is a static property of the deployment that is fully known at build time.

## Solution

Resolve the deployment shape **exactly once, at build time**, with a real PSL (`tldts`), and **bake the
result in as a literal** that both bundles read at runtime. Runtime code branches on a constant; no
runtime module ever imports `tldts`; the client bundle and the api bundle both ship zero PSL code.

The proven runtime behavior of #720 (the cross-origin session handoff, the nonce binding, the
`skipStateCookieCheck` split path, the `SameSite=None` split cookies) is **carried over intact** -
it is verified and must not be redesigned. What changes is (a) _how the mode is detected_ (build-time
`tldts` instead of runtime string-match), (b) _how wide the split trigger is_ (any pair that cannot
share a cookie domain, not only "api sits on a public suffix"), and (c) _where the code lives_
(`packages/env` owns detection + the baked value; `packages/auth` owns the runtime behavior;
`packages/config` hosts none of it).

The operator experience is unchanged: same env vars, no new required config, no secrets on the web app,
works on free-tier suffixes and custom domains alike, one deploy shape.

---

## Locked decisions

These were decided with the maintainer and are **not open** for the spec to relitigate:

- **L1. Placement: env = value, auth = behavior.** `packages/env` owns the build-time detector and
  exposes the baked `DEPLOY_MODE` as one more validated env value (its exact charter). `packages/auth`
  owns all runtime behavior keyed on that value and the handoff mechanics. `packages/config` hosts
  **no** deploy logic (it stays clean - on `canary` `config/src` is only `site.ts`, and it must stay
  that way; the `deploy.ts` #720 added to config is not ported).
- **L2. Trigger width: wide.** The handoff activates whenever the web origin and the api origin
  **cannot share a cookie domain** (they have no common registrable domain, per `tldts`), which
  includes both "api on a public suffix" _and_ "two different custom domains on different parents".
  #720's narrower "api sits on a public suffix" trigger only boot-warned about the two-custom-domain
  case; the wide trigger closes that gap. Cost is ~zero because we already run `tldts` at build.
  Caveat to record in the PR: only the public-suffix pair is field-verified; the two-custom-domain
  arm is logically correct but not yet live-tested (see Testing).
- **L3. `tldts` is a dev-only dependency, everywhere it appears, and must never enter a shipped
  bundle.** This is the north-star invariant and must be enforced by an automated check (see Testing).
- **L4. Dev is never split.** Portless local dev serves web and api as subdomains of one
  `.localhost` parent (shared-domain). `resolveDeployMode` short-circuits `localhost`/`.localhost`
  hosts to `shared` without consulting `tldts`. Dev therefore never needs `tldts` and never takes the
  split path.

---

## User stories

1. As an operator deploying the template as two `*.vercel.app` projects, I want sign-in to work on my
   first deploy with no extra configuration, so that the "deploy in five minutes" promise holds without
   a custom domain.
2. As an operator deploying to two different custom domains (e.g. `app.acme.com` + `api.othervendor.io`),
   I want sign-in to work, so that a split-vendor topology is not a silent auth failure.
3. As an operator on a shared domain or subdomains of one domain, I want the exact same behavior and
   performance as today, so that the common path is provably a zero-regression.
4. As an end user, I want to sign in successfully in Safari and Firefox (not only Chrome), so that
   cross-site-cookie-blocking browsers are not broken.
5. As a maintainer, I never want to hand-edit a list of hosting suffixes again, so that a newly launched
   platform suffix does not silently break detection.
6. As a maintainer, I want zero PSL code in either shipped bundle and no per-request PSL work, so that
   the correctness win carries no runtime weight.
7. As a maintainer, I want the deploy shape decided in one place at build and read as a constant
   everywhere, so that the behavior is auditable and cannot drift between the client and the api.
8. As an AI implementer, I want the verified handoff/nonce/`skipStateCookieCheck` behavior specified as
   "port, do not redesign", so that I preserve the field-tested security posture exactly.

---

## Implementation decisions

> **AMENDED during implementation (maintainer call): bake facts, not decisions.** The api build bakes
> only the Public Suffix List facts (`COOKIE_DOMAIN` + `COOKIE_DOMAIN_PUBLIC_SUFFIX`, resolved by
> `@packages/env/deploy`); the same-or-cross decision runs at boot in `@packages/auth/deploy` (pure,
> client-safe), so an env change takes effect on restart with no rebuild, and the baked-decision
> mismatch/downgrade machinery this spec described became unnecessary. The web build bakes the
> decision itself (`NEXT_PUBLIC_DEPLOY_MODE`) using the same `resolveDeployMode` function, since the
> browser has no runtime env. Two deliberate narrowings of the pre-amendment prose, both test-pinned:
> "shared" requires the api's environment-scoped parent (canary getCookieDomain semantics, preserving
> the #715 cookie isolation) to cover a trusted web origin, not merely a common registrable domain, so
> an env-scope mismatch or an apex api resolves split (which works) instead of a wider cookie; and the
> mode names are `host-only | shared | split`. Testing item 3's letter ("auth branches on
> env.DEPLOY_MODE") maps onto the decision-function + option-plumbing tests; asserting the assembled
> Better Auth instance applies split config end to end needs the api test harness already iceboxed in
> `handoff-route-tests.md`. The sections that follow are the pre-amendment design, kept as the trail.

### Current state on `canary` (the base this builds on)

- `packages/env/src/`: per-consumer entries `web-next.ts`, `api-hono.ts`, `auth.ts`, `db.ts`, an
  `index.ts` (exports `getSafeEnv`), and `lib/`. Runtime env is validated per consumer.
- `packages/auth/src/`: `index.ts` (the Better Auth instance) and `lib/utils.ts` (currently
  `getCookieDomain`/`getCookiePrefix`, including the `.localhost` carve-out from #715). No deploy-mode
  tri-state on canary; that is a #720 concept being reintroduced build-time-native here.
- `packages/config/src/`: only `site.ts`. The `tsdown` factory `definePackageConfig` is exported from
  `@packages/config/tsdown` and is what `api/hono/tsdown.config.ts` calls.
- `web/next/next.config.ts` already imports `@packages/env` (`getSafeEnv`, `env`).
- `api/hono/tsdown.config.ts` already imports `@packages/env` and calls `definePackageConfig`.

None of #720's split code (`config/src/deploy.ts`, `api/.../handoff.ts`, `web/.../api/handoff/route.ts`,
the `resolveDeployMode` tri-state in auth) exists on canary. The verified versions of those files live
on the #720 branch and are the **source to port from** (see "Port, replace, new" below).

### The deploy mode

One canonical tri-state enum, defined once and shared by type across build and runtime:

```
DeployMode = "shared" | "host-only" | "split"
```

- **shared** - web and api share a registrable domain (custom domain + subdomains, or dev `.localhost`).
  Behavior: today's `crossSubDomainCookies` path, byte-identical, `Lax` untouched.
- **host-only** - a bare host / unresolvable shape where no cookie domain can be asserted. Today's
  fallback, unchanged.
- **split** - web and api cannot share a cookie domain (no common registrable domain; includes
  "api on a public suffix" and "two different custom domains"). Behavior: host-only `SameSite=None`
  cookies + the nonce-bound single-use session handoff + `account.skipStateCookieCheck: true`.

### L1 concrete split (the proposed env/auth division)

**`packages/env` - owns detection + the value.**

- New **build-only** module `packages/env/src/deploy.ts`, exporting:
  - `type DeployMode` (canonical; re-exported for runtime typing).
  - `resolveDeployMode(webUrl: string, apiUrl: string): DeployMode` - pure, uses `tldts`
    (`getDomain` / `getPublicSuffix`):
    - either host is `localhost` or a `*.localhost` host → `shared` (dev; `tldts` not consulted). [L4]
    - both resolve to the **same** registrable domain → `shared`.
    - they resolve to **different** registrable domains, or either host sits **directly on a public
      suffix** (no registrable domain) → `split`. [L2, wide]
    - a URL is unparseable / a bare host with no suffix → `host-only`.
  - `resolveSharedCookieDomain(webUrl: string): string | undefined` - the `.registrable.tld`
    (or `.zerostarter.localhost` in dev) to bake for shared mode, so runtime never derives it from the
    request host with PSL logic.
- `tldts` is added as a **devDependency of `packages/env`**.
- **Isolation rule:** `deploy.ts` is imported **only** by build tooling (`next.config.ts`,
  `tsdown.config.ts`). No runtime entry (`web-next.ts`, `api-hono.ts`, `auth.ts`, `index.ts`) may import
  it, so `tldts` is never in a runtime module graph. [L3]
- **Runtime exposure:** `web-next.ts` validates `NEXT_PUBLIC_DEPLOY_MODE` (enum) and, if used,
  `NEXT_PUBLIC_COOKIE_DOMAIN`; `api-hono.ts` validates `DEPLOY_MODE` and `COOKIE_DOMAIN`. Both expose
  the typed `DeployMode` value the same way every other env value is exposed today. Runtime code reads
  `env.DEPLOY_MODE`, never recomputes it.

**`packages/auth` - owns behavior.**

- Consolidate the runtime handoff mechanics here (moved out of where #720 put them in config):
  `mintHandoffToken`, `HANDOFF_TOKEN_PATTERN`, `HANDOFF_NONCE_COOKIE`, the `handoffIdentifier(id, nonce)`
  helper, next to the existing `getCookieDomain`/`getCookiePrefix` in `lib/utils.ts`.
- **Delete** #720's runtime `resolveDeployMode` (the string-suffix version). Auth reads
  `env.DEPLOY_MODE` once at module init and branches: `shared` → `crossSubDomainCookies` with the baked
  `COOKIE_DOMAIN`; `host-only` → fallback; `split` → `SameSite=None` host-only + handoff +
  `account.skipStateCookieCheck: true`.
- No `tldts` in auth's runtime graph. `getCookieDomain` returns the baked shared cookie domain (or the
  dev `.localhost` value) with no PSL parsing.

**`packages/config` - build plumbing only, no deploy logic.**

- `definePackageConfig` (`@packages/config/tsdown`) gains a **generic** `define` pass-through so the api
  build can bake a literal. This stays generic (it is not deploy-aware); the deploy value is computed in
  `env` and handed to the factory by the api's `tsdown.config.ts`.

### Build-time injection

- **Web** (`web/next/next.config.ts`, already imports env): compute
  `resolveDeployMode(webUrl, apiUrl)` from `@packages/env/deploy`, then expose it to the client as
  `NEXT_PUBLIC_DEPLOY_MODE` (and `NEXT_PUBLIC_COOKIE_DOMAIN` if used) via Next's `env` config /
  `process.env` so Next **inlines the literal** into the client bundle. `tldts` runs here only, at build.
- **API** (`api/hono/tsdown.config.ts`): compute the mode from `@packages/env/deploy`, pass
  `define: { "process.env.DEPLOY_MODE": JSON.stringify(mode), ... }` through `definePackageConfig`.
  `tsdown` bakes the literal; the non-taken mode branches and the `tldts` import **tree-shake out** of
  the api bundle. `tldts` runs here only, at build.
- The URLs to feed `resolveDeployMode` come from the same env the apps already read
  (web origin + `NEXT_PUBLIC_API_URL` / `HONO_APP_URL`); no new required env var is introduced.

### Port, replace, new

**Port from the #720 branch verbatim in behavior (verified - do not redesign):**

- The api handoff router (`/start` + `/claim`, mode-gated to 404 outside split, the 64-hex token regex,
  the `handoffIdentifier` nonce-in-identifier scheme, single-use `createVerificationValue` /
  `consumeVerificationValue`, 60s TTL).
- The web claim route (mode-gated 404, the `maxAge` `Number.isFinite` guard, `fail()` keeping the nonce
  on failure, first-party cookie set on success).
- The per-sign-in nonce minted as a first-party web cookie and threaded through the handoff.
- `account.skipStateCookieCheck: true` in split mode - **A/B-proven load-bearing on Safari** (P0 true →
  dashboard; P1 false → breaks at the api origin). Carries the documented, accepted login-CSRF tradeoff
  (see Further notes).
- `SameSite=None` host-only cookies in split mode.
- `getSession` forwarding cookie only (no `x-forwarded-for`) + `cache: "no-store"`.

**Replace (the point of this PR):**

- Detection: runtime string-match against `PUBLIC_HOSTING_SUFFIXES` → build-time `tldts`. Delete
  `PUBLIC_HOSTING_SUFFIXES` entirely.
- Trigger: narrow (api-on-public-suffix) → wide (no shared cookie domain). [L2]
- Home: `config/src/deploy.ts` → split across `env` (detection + value) and `auth` (mechanics). [L1]
- Mode source at runtime: module-init recomputation → read the baked `env.DEPLOY_MODE` literal.
- Shared-mode cookie domain: runtime PSL-ish derivation → baked `COOKIE_DOMAIN`.

**New (did not exist in #720):**

- The `env/src/deploy.ts` build-only detector and the `tldts` devDependency.
- The build-injection wiring in `next.config.ts` and `tsdown.config.ts` (+ the generic `define`
  pass-through in `definePackageConfig`).
- The automated "no `tldts` in any shipped bundle" assertion. [L3]

---

## Testing decisions

**Seams (highest first, fewest possible).** The entire _novel_ correctness surface collapses to one
seam: `resolveDeployMode(web, api)` is a pure function, so a single input to output table validates every
mode decision (shared / host-only / split, the wide trigger, the dev-localhost short-circuit) at the
highest point in the stack, above build injection and above runtime branching. Everything else is a
_reused_ seam or a build gate, not a new one: the ported handoff/cookie mechanics keep #720's existing
`packages/auth/test/` seam, and the L3 "no `tldts` in a shipped bundle" invariant is a build-output
assertion that cannot fold into a unit seam. Target shape: **one new behavioral seam, one reused seam,
one build gate, one manual release gate** (below, in that order).

A good test here asserts only external behavior: a topology pair in, a mode out; a host in, a cookie
domain out; a built artifact, a grep result. None of them reaches into how detection is computed.

Prefer the existing seams. #720's tests live in `packages/auth/test/` (a `utils.test.ts` and a
`deploy.test.ts`, 14 tests total; the deploy test was deliberately relocated to `packages/auth` because
`bun:test` does not resolve under `@packages/config`). Keep tests where their subject now lives.

1. **Detection truth table (unit, in `packages/env`).** `resolveDeployMode(web, api)` over a table that
   must include, at minimum: two `*.vercel.app` projects → `split`; two subdomains of one custom domain →
   `shared`; two **different** custom domains → `split` (the wide arm, the new case); `.localhost`
   subdomains → `shared` (dev, no `tldts`); a bare host → `host-only`; and one representative row per
   public suffix shape you care about (`netlify.app`, `pages.dev`, `workers.dev`, `up.railway.app`).
   This replaces #720's `isPublicHostingSuffix`/`isSplitPair`/curated-list tests. The test lives with its
   subject in `packages/env/test/`; make `bun:test` resolve there the way `packages/auth` already does
   (add `@types/bun` to `packages/env` if absent), and do **not** edit the shared `@packages/config`
   tsconfig to force resolution (standing rule from the #720 review).
2. **Handoff + cookie mechanics (unit, in `packages/auth`).** `getCookieDomain` table (including the
   `.localhost` carve-out and the baked shared domain), `mintHandoffToken` shape, `HANDOFF_TOKEN_PATTERN`,
   `handoffIdentifier`. Ported from #720.
3. **Runtime reads the baked literal, not a recomputation.** A test proving auth branches on
   `env.DEPLOY_MODE` and that no runtime module imports `@packages/env/deploy` (grep/import-graph
   assertion), so `tldts` cannot reach a runtime graph.
4. **The L3 invariant - no PSL in shipped bundles (the load-bearing new check).** AMENDED during
   implementation (maintainer call): enforced at the source seam, not the built artifact. A repo-wide
   test asserts no file under any `src/` imports `@packages/env/deploy` or `tldts` (import statements
   only); build configs live outside `src/`, so the boundary is exact, and bundlers cannot pull PSL
   code into an artifact no source file imports. This replaced a per-build bundle grep, which was
   dropped because the enforcement belongs in one CI test, not a step in three build scripts (and a
   name-grep is unreliable anyway: "tldts" does not survive bundling+minification).
5. **`skipStateCookieCheck` reaches `oauthConfig` at runtime.** Reuse #720's `betterAuth` +
   `memoryAdapter` test proving `oauthConfig.skipStateCookieCheck === true` in split mode on
   `better-auth@1.6.23`.
6. **Field verification (manual, required before merge).** Re-run the studiozero (`*.vercel.app` pair)
   Safari end-to-end: GitHub sign-in in a Private window with Prevent Cross-Site Tracking on lands on the
   server-gated `/dashboard`; handoff claim is single-use (replay dead); `/dashboard` is 200 with the
   handed-off cookie and 307 without. The two-different-custom-domains arm (L2) is logically covered by
   test 1 but is **not** field-verified; note that explicitly in the PR.

---

## Out of scope

Carried from #720's "Not in this PR" plus this PR's own deferrals. Track under Icebox #707.

- The full `vercel.mdx` "without a custom domain" walkthrough (this PR ships only the split-mode boot
  warning + the short note that the two-project `*.vercel.app` setup works out of the box).
- Vercel preview self-wiring (a split preview pair auto-configuring from the branch-URL pattern);
  deferred to the #677 plan. Feature-branch previews still point at explicit envs.
- Per-client rate-limit keying for the server-to-server handoff call (the api keys under the web
  function's egress IP; `getSession` stays cookie-only).
- `INTERNAL_API_URL` → `DOCKER_INTERNAL_API_URL` rename (orthogonal).
- A web-origin session proxy for Safari/Firefox _client-side_ session reads (sign-in itself is fixed;
  this is only for browser JS reading the session cross-origin).
- The tighter callback-to-nonce binding that would close the accepted split-mode login-CSRF tradeoff.
- Re-extending the split-mode web session cookie as the session rolls.
- The #719 WebSocket escape hatch (`NEXT_PUBLIC_API_WS_URL` + reconnect backoff).
- Renaming `HONO_APP_URL` → `BETTER_AUTH_URL`/`API_URL` (its own breaking change; not bundled here).

## Further notes

- **Better Auth's own position (research verdict).** There is no first-class Better Auth fix for two
  `*.vercel.app` projects - the maintainer (ping-maxwell, discussion #5073) states it is "not something
  Better-Auth can fix... browser-level cookie isolation via `.vercel.app`", and recommends: a reverse
  proxy (= the closed Plan B / PR #716), a custom domain + `crossSubDomainCookies`, or OIDC/Bearer.
  Issues #4270 / #4878 are closed-not-planned; #7283 is open-unanswered. Our handoff is Better Auth's own
  One-Time-Token "SSO handoff" pattern, nonce-hardened; `oauth-proxy` is itself a redirect handoff and
  takes the same `state`-cookie posture we take. This PR fills an acknowledged gap with sanctioned
  primitives; it does not fight the framework.
- **Accepted split-mode tradeoff (documented at the call site).** `skipStateCookieCheck` validates the
  OAuth callback against the DB state (single-use, server-issued CSPRNG, deleted on parse) but not a
  browser-bound cookie, so a relayed callback can set the api-origin session in another browser - a
  bounded login-CSRF (mostly Chrome third-party cookies; SSR stays protected by the nonce). Inherent to
  cookie-less cross-origin OAuth and the posture `oauth-proxy` takes. The tighter binding that closes it
  is deferred (see Out of scope, Icebox #707).
- **Zero-regression claim, unchanged from #720.** One operator config shape everywhere; no new required
  env var; no secrets on the web app; the mode is a static literal so the per-request path gains no
  instructions in any mode; split-mode cost is one extra redirect + one insert/delete per sign-in only.
  The added correctness (real PSL, wide trigger) carries no runtime weight because it is all build-time.
- **Naming.** Deliberately not locked in this document. Lock the names (the `DeployMode` member spellings,
  `resolveDeployMode` vs alternatives, `HANDOFF_*` constants, the `DEPLOY_MODE` env key) as the first
  step of the spec, before code, per the maintainer's "talk naming from the start". The names used above
  are descriptive placeholders, not decisions.

nrjdalal:
Closes #721. Part of #719; supersedes the detection half of #720 (the verified handoff behavior is ported intact).

Cookie auth breaks when the web and api apps deploy as two origins that cannot share a cookie: two projects on a public hosting suffix (plain `vercel.app`, `netlify.app`, and friends), or two unrelated custom domains. The OAuth state cookie is rejected at storage and sign-in dies with `state_mismatch` on the template's first-touch deployment path.

## What this does

**Bakes facts, not decisions.** The api build resolves the one question only the Public Suffix List can answer, the api's parent cookie domain and whether browsers reject a `Domain` cookie there, via `tldts` (a dev-only dependency of `@packages/env`, consumed exclusively by the build-only `@packages/env/deploy`), and bakes the answer as two literals: `COOKIE_DOMAIN` + `COOKIE_DOMAIN_PUBLIC_SUFFIX`. The same-or-cross decision itself runs at boot in `@packages/auth/deploy` (pure, client-safe, no env, no PSL) from those facts plus `HONO_TRUSTED_ORIGINS` (the web counterpart is the first trusted origin on a different host; a boot warning flags ambiguity when several are listed), so an env change takes effect on restart with no rebuild. The web build runs the exact same decision function at build and inlines `NEXT_PUBLIC_DEPLOY_MODE` (browsers have no runtime env), so the client ships a literal instead of the suffix predicate #720 shipped to the browser.

The decision is the tri-state `DeployMode`:

- **shared**: a usable parent domain covers a trusted web origin (custom domains, portless localhost). Today's `crossSubDomainCookies` path, byte-identical, `Lax` untouched, environment-scoped domains (#715) preserved.
- **host-only**: no distinct web origin (plain-localhost dev, api-only deploys, one site). Today's fallback.
- **split**: any remaining pair of distinct hosts, now including two unrelated custom domains and env-scope mismatches (the wide trigger #720 only boot-warned about). Host-only `SameSite=None` cookies plus the nonce-bound single-use session handoff, ported verbatim in behavior from the field-verified #720 branch: `/start` parks the session cookie value under a one-time id via Better Auth's own verification adapter (60s TTL, atomic single-use consume, nonce folded into the row identifier), the web claims it server-to-server and sets the cookie first-party, and both handoff surfaces answer 404 as if absent outside split mode. Split mode also keeps `account.skipStateCookieCheck` (A/B-proven load-bearing for Safari on the live pair; the DB-persisted single-use `state` stays the CSRF binding, the posture Better Auth's own `oauth-proxy` takes; accepted tradeoff documented at the call site and in `plans/split-oauth-callback-binding.md`).

**Why the PSL is finally real.** #720 hand-curated `PUBLIC_HOSTING_SUFFIXES` (12 entries, stale the day a new platform launches) because a PSL library was too heavy for the client and the hot path. Baking at build removes both objections: `tldts` never ships (a repo-wide isolation test pins that no file under any `src/` imports it or the build-only module, with the scan roots derived from the workspace globs so a new package is swept automatically; bundlers cannot pull in what no source imports), and the runtime does zero PSL work in any mode. The curated set, `isSplitPair`, and `isPublicHostingSuffix` are gone; `packages/config` hosts no deploy logic.

## Zero regressions, by construction

- One operator config shape everywhere: same URL vars, no new required env, no secrets on the web app. `COOKIE_DOMAIN`/`COOKIE_DOMAIN_PUBLIC_SUFFIX`/`NEXT_PUBLIC_DEPLOY_MODE` are auto-derived, never hand-set, deliberately absent from `.env.example` (documented in `environment.mdx`, mirrored in turbo `globalEnv` per the schema-mirror convention).
- Shared-domain deployments run byte-identical cookie config; the per-request path gains no instructions in any mode. Split-mode cost is one extra redirect and one insert plus delete, per sign-in only.
- Dev never depends on `tldts` and never resolves split: `.localhost`/`localhost` short-circuit before the PSL, and an unbaked from-source run falls back to the exact pre-split derivation.
- 20 unit tests across the three seams: the PSL-fact table (`resolveCookieDomain`), the same-or-cross truth table (`resolveDeployMode`, including every wide-trigger row and canary-parity row), the handoff token shape, the `skipStateCookieCheck` plumbing pinned against `better-auth@1.6.23`, and the repo-wide isolation pin.

## Verification

- Full monorepo: check-types green, 116 tests green, api + web builds green.
- Two-axis review (standards + spec vs #721) ran pre-push; findings closed in the `fix(review)` commit: workspace-derived isolation-scan roots, turbo `globalEnv` mirror, stale `DEPLOY_MODE` doc wording, em-dash strip, and the spec amendment now records the deliberate narrowing (shared requires the api's environment-scoped parent to cover the web, preserving #715 cookie isolation; an env-scope mismatch or apex api resolves split, which works, instead of a wider cookie).
- Baked artifacts inspected: the api bundle carries the `COOKIE_DOMAIN*` literals (no `process.env` read survives), the client chunk carries `NEXT_PUBLIC_DEPLOY_MODE:"host-only"` as a plain literal, and a byte-scan of `.next` + the api bundle finds zero PSL code (the only "tldts" matches in build output are this PR's own documentation prose).
- Local e2e (dev = shared mode, the zero-regression surface): agent sign-in through the real browser lands on the server-gated `/dashboard`; the handoff routes 404 as designed. Login dialog: https://litter.catbox.moe/tpr7pg.png Signed-in dashboard: https://litter.catbox.moe/oa5ec8.png
- **Pending before merge:** the split-mode field re-run on the live `*.vercel.app` pair (studiozero), including the Safari sign-in. #720 verified the identical handoff behavior end to end there; this PR moves detection, not the flow, but the re-run should still happen on this exact diff. The two-custom-domains wide arm is unit-covered but has no live testbed; flagged per the spec.

## Not in this PR

Unchanged from #720's deferrals, tracked in Icebox #707 and `plans/`: the full `vercel.mdx` no-custom-domain walkthrough, Vercel preview self-wiring (#677), per-client rate-limit keying for server-to-server calls, the A2 web-origin session proxy for client-side cross-origin session reads, the tighter callback-to-nonce binding, split-mode web cookie lifetime re-extension, the WebSocket escape hatch, and handoff route integration tests (blocked on an api test harness).

## Summary by CodeRabbit

* **New Features**
  * Added automatic deployment-mode detection for shared, split, and host-only setups.
  * Added secure sign-in handoff for split deployments where sites cannot share cookies.
  * Improved OAuth, magic-link, and social sign-in flows across separate web and API domains.
  * Added support for safer cookie behavior across subdomains and local/IP-based environments.

* **Bug Fixes**
  * Improved session retrieval by forwarding only relevant cookie information.

* **Documentation**
  * Expanded deployment, authentication, and environment-variable guidance for split deployments and custom domains.


nrjdalal:
Superseded by #727 (dissolve, merged to canary as 4f4922f4) plus #723 (build-time tldts cookie facts, already on canary).

This is the teleport/handoff design: the same nonce park/claim, `SameSite=None`, and `skipStateCookieCheck` ported from #720. #727 replaced the handoff with routing sign-in same-origin through the web's `/api` proxy, and #723 already bakes the PSL cookie facts, so both halves here are now redundant (and it conflicts heavily with canary).

The one capability dissolve doesn't yet cover, split for **two unrelated custom domains / env-scope mismatches** (dissolve triggers only on a public hosting suffix), is deferred to the Icebox (#707) as a widen-the-trigger follow-up, not a reason to keep the handoff. #721's public-suffix case is covered by #727. Closing.

PR #723, opened 2026-07-17
feat(auth): derive cookie config from a build-generated tldts breakdown (@packages/scripts)

nrjdalal:
## What

Derive the session-cookie configuration from a real tldts parse of the app host, run **at build time in a dedicated scripts package** so its Public Suffix List never ships in the runtime bundle.

- **`@packages/scripts`** — a new private workspace package for app-build tooling (never imported at runtime, so never bundled). It mirrors the standard `packages/*` shape: source under `src/`, `tsconfig.json` extends `@packages/config/tsconfig.json`, and its build-only deps are devDependencies (`@packages/config`, `@packages/env`, `tldts`, off the root). `src/generate-tldts.ts` reads `env.HONO_APP_URL` from `@packages/env` (reusing its env loading, no separate dotenv), runs `parse(url, { allowPrivateDomains: true })`, and writes the repo-root `.generated/tldts.json` (gitignored). Generated assets get one centralized `.generated/` home (dot-prefixed like `.next`/`.source`/`.turbo`; "generated" not `.cache` since it's a required build input, not a perf cache), which `.gitignore`/`.dockerignore` and `bun run clean` all cover. `.github/scripts` stays for CI/repo tooling.
- **`@packages/auth` build runs it first** (`"build": "bun ../scripts/src/generate-tldts.ts && tsdown"`, with `@packages/scripts` as a devDep so `turbo prune` keeps it in Docker). `tsdown.config.ts` reads the JSON and bakes it via `define` (`definePackageConfig` gained an optional `define` passthrough). `index.ts` reads the ambient constant and runs one `cookieConfig(tldts)` → `{ cookieDomain, cookiePrefix, isPrivate }`, switching cookie mode on `isPrivate`.
- **`@packages/env` stays plain `createEnv(...)`** — the earlier `derived` pass-through is gone, since auth is the only consumer.

## Why the full PSL, at build

`co.uk`, `com.au`, `s3.amazonaws.com`, `vercel.app` and friends can't be handled by a small hand-rolled list — correct public-suffix boundaries need the PSL. Running it at build gets that correctness (e.g. `api.example.co.uk` → `cookieDomain: ".example.co.uk"`, `cookiePrefix: undefined`, not a bogus `"example"`) while shipping only the ~200-byte result. Tree-shaking can't help: the PSL is data a called `parse()` consumes, not dead code.

## Cookie mode

`index.ts` `advanced` branches on `isPrivate` (tldts's own field, passed through):

- **hosting suffix** (`vercel.app`, `pages.dev`, `github.io` → `true`): sibling origins can't share a Domain cookie → `defaultCookieAttributes: { sameSite: "none", secure: true }`.
- **otherwise**: `crossSubDomainCookies` scoped to the derived `cookieDomain`.

## Behavior

Custom-domain and `.localhost` hosts unchanged. Hosting suffixes → host-only + `SameSite=None` (canary emitted the browser-rejected `.vercel.app`). ccTLD apex → host-only; any IP literal → host-only. A host with no shareable parent (apex API URL, bare `localhost`, IP) keeps the `cookieDomain &&` guard, staying host-only rather than widening to `Domain=<hostname>`.

## Build size

PSL is build-only, so the runtime bundle is essentially unchanged from canary:

| Artifact | canary | this PR |
|---|---|---|
| `@packages/env/dist/auth.mjs` | ~1.0 KB | ~1.0 KB (unchanged) |
| `@packages/auth/dist/index.mjs` | ~1.8 KB | ~2.0 KB (baked breakdown) |
| `@api/hono` deployed bundle | 1.43 MB | 1.43 MB |
| `@web/next` | — | unchanged |

## Build-time trade-off

The cookie config is baked to the **build-time** `HONO_APP_URL`, the same way the web bakes `NEXT_PUBLIC_*`. The script sources it as the runtime does (`process.env` on Vercel, the mounted `.env` on Docker), and `turbo.json` re-hashes the build on a URL change via `globalEnv` (`HONO_APP_URL` from `process.env`) plus `.env*` in `globalDependencies` (for the `.env`/`.env.<mode>` file case that `globalEnv` can't see). Docker layer caching is separate: BuildKit excludes the mounted `.env` secret, so a `.env` change needs `docker compose build --no-cache` (documented in the Docker guide). Running one image against a *different* runtime `HONO_APP_URL` would update `baseURL` but not the baked cookie config; the standard flows (Vercel per-deploy, docker-compose sharing one `.env`) match.

## Verification

- Full turbo build (all 8 packages) + `check-types` (incl. `@packages/scripts`) green; tldts absent from every `dist` and the api bundle.
- Script sourcing verified: with `HONO_APP_URL` only in a `.env` file (shell var unset), the script wrote the correct `api.zerostarter.dev` breakdown and it baked into `auth/dist`.
- `packages/auth/test/utils.test.ts` covers `cookieConfig` across production subdomains, `.localhost` (main + worktree), ccTLD apex/subdomain, hosting suffixes, IPs, and the null-host fallback (`bun test` green).
- Since this adds a workspace package, `@packages/scripts` conforms to the standard `packages/*` shape (`src/` layout, `tsconfig.json` extends `@packages/config/tsconfig.json`), and a new `add-package` skill (`.agents/skills/add-package`, registered in the AGENTS.md table) captures that shape for the next one.
- Working notes: `.github/notes/plans/env-derived-tldts.md`.

## Scope

canary only. Independent of the `feat/split-deploy-auth` branch — the `isPrivate` signal is computed straight from tldts's PSL, not that branch's curated set.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved authentication cookie domain and prefix handling across localhost, IP addresses, subdomains, and multi-level domains.
  * Improved environment-specific cookie isolation while preserving correct behavior for production domains.

* **Tests**
  * Expanded coverage for cookie configuration across public suffixes, localhost, IP addresses, and nested subdomains.

* **Documentation**
  * Added planning notes describing the updated authentication cookie handling.

### DIFF
--- a/packages/scripts/package.json
+++ b/packages/scripts/package.json
@@ -1,0 +1,17 @@
+{
+  "name": "@packages/scripts",
+  "version": "0.0.0",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "check-types": "tsc --noEmit"
+  },
+  "devDependencies": {
+    "@packages/config": "workspace:*",
+    "@packages/env": "workspace:*",
+    "@types/bun": "catalog:",
+    "@types/node": "catalog:",
+    "tldts": "catalog:",
+    "typescript": "catalog:"
+  }
+}
--- a/.agents/skills/add-package/SKILL.md
+++ b/.agents/skills/add-package/SKILL.md
@@ -1,0 +1,139 @@
+---
+name: add-package
+description: Add a new shared workspace package under packages/*. Use when creating a new @packages/<name>, whether a bundled library other workspaces import or a build-only script package that runs during a build.
+---
+
+# Add a Package
+
+Workspaces are globbed as `api/*`, `packages/*`, `web/*` (root `package.json`). A new shared package lives at `packages/<name>/` and is named `@packages/<name>`. Every package shares one shape; copy an existing sibling, do not invent a new layout.
+
+## Pick the shape
+
+- **Library** (`env`, `db`, `auth`, `config`): built with tsdown to `dist/`, imported by other workspaces via an `exports` map. Use when code is consumed at runtime.
+- **Build-only script** (`scripts`): never bundled, never imported at runtime. Its `.ts` files run via `bun src/<x>.ts` during another package's build (e.g. `@packages/auth`'s `build` runs `bun ../scripts/src/generate-tldts.ts` first). Use for build-time codegen. Keep CI/repo tooling in `.github/scripts` instead; `packages/scripts` is for app-build tooling that needs workspace deps.
+
+## Common skeleton (both shapes)
+
+```
+packages/<name>/
+├── package.json
+├── tsconfig.json
+└── src/
+    └── index.ts        # source under src/, imported via @/ (a build-only script names its entry by function, e.g. generate-tldts.ts, not index.ts)
+```
+
+`package.json` fields shared by every package:
+
+```jsonc
+{
+  "name": "@packages/<name>",
+  "version": "0.0.0",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "check-types": "tsc --noEmit"
+  },
+  "devDependencies": {
+    "@packages/config": "workspace:*",
+    "@types/bun": "catalog:",
+    "@types/node": "catalog:",
+    "typescript": "catalog:"
+  }
+}
+```
+
+`@packages/config` MUST be a devDependency: `tsconfig.json` extends it, and without the dep the `extends` cannot resolve. Keep deps and `exports` alphabetical (A→Z); catalog-versioned deps use `"catalog:"`, workspace deps use `"workspace:*"`.
+
+`tsconfig.json` (identical to `env`/`db`/`auth`):
+
+```json
+{
+  "extends": "@packages/config/tsconfig.json",
+  "compilerOptions": {
+    "paths": {
+      "@/*": ["./src/*"]
+    }
+  },
+  "include": ["**/*.ts", "**/*.tsx"],
+  "exclude": ["node_modules"]
+}
+```
+
+## Library shape (adds to the skeleton)
+
+```jsonc
+{
+  // ...skeleton...
+  "files": ["dist"],
+  "exports": {
+    ".": { "types": "./dist/index.d.mts", "default": "./dist/index.mjs" }
+  },
+  "scripts": {
+    "build": "tsdown",
+    "check-types": "tsc --noEmit"
+  },
+  "dependencies": { /* runtime deps */ },
+  "devDependencies": {
+    "@packages/config": "workspace:*",
+    "@types/bun": "catalog:",
+    "@types/node": "catalog:",
+    "tsdown": "catalog:",
+    "typescript": "catalog:"
+  }
+}
+```
+
+Add one export entry per `entry` file. `tsdown.config.ts` uses the shared helper (validates env in `build:prepare`, emits tsgo dts, minifies):
+
+```ts
+import { definePackageConfig } from "@packages/config/tsdown"
+import { getSafeEnv } from "@packages/env"
+import { env } from "@packages/env/<name>"
+
+export default definePackageConfig({
+  name: "@packages/<name>",
+  env,
+  getSafeEnv,
+})
+```
+
+A package with no env of its own can pass another package's `env`/`getSafeEnv`, or (like `env` itself) call `defineConfig` from tsdown directly. `turbo.json`'s `build.outputs` already lists `dist/**`, so no turbo edit is needed.
+
+## Build-only script shape (adds to the skeleton)
+
+No `build`, no `exports`, no `files`, no `tsdown`. Add only the script's own tool deps (e.g. `tldts`) as devDependencies. Because the entry is a Bun script using `Bun.*` / `import.meta.dir` / `node:*`, the native tsc preview (tsgo) does not auto-include `@types/*` for it, so pin `types: ["bun"]` exactly as `packages/cli` (the repo's other Bun package) and `.github/scripts/tsconfig.json` do:
+
+```json
+{
+  "extends": "@packages/config/tsconfig.json",
+  "compilerOptions": {
+    "types": ["bun"],
+    "paths": {
+      "@/*": ["./src/*"]
+    }
+  },
+  "include": ["**/*.ts", "**/*.tsx"],
+  "exclude": ["node_modules"]
+}
+```
+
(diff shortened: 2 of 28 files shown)


packages/scripts/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "@packages/scripts",
+  "version": "0.0.0",
+  "private": true,
+  "type": "module",
+  "scripts": {
+    "check-types": "tsc --noEmit"
+  },
+  "dependencies": {
+    "dotenv": "catalog:",
```
nrjdalal: should have used `@packages/env`

nrjdalal: Done in d648946d: \`generate-tldts\` now imports \`env.HONO_APP_URL\` from \`@packages/env\` (reusing its env loading) instead of its own dotenv, and \`@packages/scripts\` deps (\`@packages/env\`, \`tldts\`) are devDependencies since it is build-only tooling. Verified the build + 13/13 check-types pass; api bundle stays 1.43 MB.

nrjdalal:
Thanks for the thorough pass. Addressed in `63edb85b` (plus what earlier commits already covered):

**Blocking**

1. **Empty CI test task** — re-added `packages/auth/test/utils.test.ts` and the `"test": "bun test"` script (`bun test` green: prod subdomains, `.localhost` main + worktree, ccTLD apex/subdomain, hosting suffixes, IP, and the null-host fallback). This reverses the earlier `73aef671`/`07f80cf3` deletions, which I'd made on an explicit "delete the tests" instruction; your recommendation (fix CI by testing `cookieConfig`, not by dropping the script) is the right call and takes precedence.
2. **Dropped `cookieDomain &&` guard widened scope** — restored. The non-suffix branch is back to `cookieDomain ? { crossSubDomainCookies } : {}`, so a host with no shareable parent (apex API URL, bare `localhost`, IP) stays host-only instead of falling back to `Domain=<hostname>`. Covered by the "host-only for apex/bare/IP/null-host" test.
3. **Silent polyfill baking a wrong config** — the loud half already landed in `d648946d`: `generate-tldts` imports `HONO_APP_URL` from `@packages/env`, which throws on a missing/invalid value rather than polyfilling. The remaining `tsdown`-without-generate escape hatch now bakes a **type-honest null host** (`{"domain":null,…,"isPrivate":null}`) → host-only, `isPrivate: null`, not an `undefined`-field object.

**Type honesty / double-stringify** — fixed together: `define` now inlines the breakdown as an **object literal** (dropped the `JSON.stringify(JSON.stringify(...))` and the runtime `JSON.parse`); `cookieConfig` takes a `ParsedHost`, and the null-host fallback keeps `isPrivate` typed `boolean | null`.

**Doc-sync + em-dash** — `@packages/scripts` added to `AGENTS.md`/`CLAUDE.md`, the `codebase-map` and `runtime-apis` skills; `authentication.mdx` cookie-scope note corrected; stale package count dropped from `README`; plan doc made canary-relative with the em-dash removed and the now-stale "dropped guard" caveat rewritten to match the restored guard.

**Left as-is (judgment calls, happy to change):**
- Duplicated `publicSuffix === "localhost"` check in `cookieConfig` — the two reads gate two independent outputs (`cookieDomain` vs `cookiePrefix`); folding them into one local read reads worse to me than the repetition.
- `packages/scripts/tsconfig.json` standalone rather than extending `@packages/config/tsconfig.json` — matches the `.github/scripts` convention; it's build-only Bun tooling with different module settings from the package `src` builds. Can extend the base if you'd prefer consistency.
- Intra-PR churn (`env.derived` added then removed) — that history is the live design exploration; squash-on-merge collapses it.

nrjdalal:
Went through the CodeRabbit findings (`483a40da`, `4c9ffe50`):

**`env.config.tldts` hallucination** — stale. That reference predates the redesign; the current `utils.ts`/`utils.test.ts` don't mention `env.config` or `derived` at all (`cookieConfig(host: ParsedHost)` now). No change needed.

**Hardcoded `../scripts/src/generate-tldts.ts` across packages** — kept, deliberately. The suggested `bin` doesn't work here: bun doesn't link workspace-package bins into `.bin` (the `zerostarter` CLI's own bin isn't linked either), so `generate-tldts` resolves to nothing and the build fails. More to the point, `bun ../<path>.ts` **is** the repo's established convention for sibling build scripts, `web/next` and `api/hono` both call `bun ../../.github/scripts/*.ts` this way. Adopting `bin` would break the build and drift from that pattern.

**Alphabetize `define`/`deps`** — done in `packages/config/tsdown.ts` (`define` before `deps` in the type, destructure, and spread; the emitted object is now A→Z). I did **not** reorder the `definePackageConfig` call in `auth/tsdown.config.ts`: its `name, env, getSafeEnv, define` is a meaningful order shared with `db` and `api` (`…, deps`), and alphabetizing just that one call would make it inconsistent with its two siblings.

**Verbose build comments** — trimmed both (`generate-tldts.ts`, `auth/tsdown.config.ts`) to a single essential line each.

Separately, generated build assets now live in one gitignored repo-root `.generated/` dir instead of inside `@packages/auth`, and `bun run clean` was brought up to date with the full generated set.

nrjdalal:
### Docker smoke (docker-test skill) — passed

Ran the `--no-cache` api build + boot against a dummy `.env` with a **real** `HONO_APP_URL=https://api.canary.zerostarter.dev`, to exercise the build path this PR adds inside a real image.

- **Build (`docker compose build --no-cache api`)** — green. `@packages/auth:build` ran inside the `turbo prune`d tree with `HONO_APP_URL: "https://api.canary.zerostarter.dev"` from the mounted `.env` secret, so `generate-tldts` executed, the new drift guard passed, and tsdown baked the result. Confirms `@packages/scripts` survives prune (kept via the auth devDep) and the mounted `.env` reaches the script at the pruned tree's depth.
- **Baked scope is correct end-to-end** — `bundle/index.mjs` in the image contains `crossSubDomainCookies` and `canary`, i.e. the canary host derived to `.canary.zerostarter.dev` + prefix `canary`, not the host-only fallback.
- **Boot + health** — container up, `GET /api/health` → `200 {"data":{"message":"ok","version":"0.1.5","environment":"local"}}`.
- **Offline self-containment** — booted with `--network=none`; `/api/health` served, no `Cannot find package`, no `.bun/install/cache` writes. The `auth`-into-api bundle resolves every import with no runtime auto-install.

Teardown clean. Web image not built (this PR doesn't touch web's libc pruning).

nrjdalal:
Re: the `turbo.json` `globalDependencies` comment — applied. Removed `packages/scripts/src/**`, keeping only `.env`.

Verified the workspace graph already covers the generator: with just `.env` in `globalDependencies`, editing `generate-tldts.ts` still produced a turbo **cache miss** on `@packages/auth:build` (hash `6c296828` → `33b36ebb`), because `@packages/auth` declares `@packages/scripts` as a devDependency, so its source feeds auth's build hash. The extra glob was redundant and over-broad (it busted the whole-repo cache on any scripts change).

`.env` stays: it isn't a package source file, so the graph doesn't cover it, and a `.env`-only `HONO_APP_URL` change would otherwise restore a stale baked-cookie `dist`. Thanks for the catch.

biswajeet-lightwork (approved):
Let's go 🚀

PR #724, opened 2026-07-17
chore(deps): replace tsgo with typescript 7

nrjdalal:
## Summary

TypeScript 7 ships the native compiler as the `typescript` package, so the `@typescript/native-preview` (tsgo) dev preview is now redundant. The catalog already pinned `typescript: ^7.0.2`, meaning every workspace's `tsc --noEmit` was already running the native compiler; this PR deletes the tsgo leftovers.

- Root `package.json`: swap the `@typescript/native-preview` devDependency for `typescript: catalog:` (the root previously had no `tsc` bin, which is why the type-check scripts used `tsgo`), drop the catalog entry, and point `check-types:scripts` and `check-types:tests` at `tsc`.
- tsdown configs (`packages/config` shared + own, `packages/env`, `packages/cli`): `dts: { tsgo: true }` -> `dts: true`. `rolldown-plugin-dts` auto-selects the native generator whenever the installed TypeScript major is >= 7, so the explicit opt-in was already dead config: canary logs `Emit types with typescript@7.0.2` *with* `tsgo: true` set.
- Docs and skills: `scripts.mdx` now says `tsc` (and documents `check-types:tests`, which was missing), `project-structure.mdx` and `.github/notes/dependencies.md` document the named-catalog exception below, `AGENTS.md` gains the rule, and the `add-package` skill drops its three tsgo references.
- `packages/cli/src/pkg.ts`: add `catalogs` to the fork sync's `MERGE_FIELDS`. `mergePkg` spreads the starter so the block reaches a fork either way, but without this a fork's own named catalog was clobbered on re-baseline, the one thing `catalog` is on that list to prevent.
- Tests: `tests/github/scripts/deps-manager.test.ts` (new) and a `catalogs` case in `tests/packages/cli/src/pkg.test.ts`.
- `bun.lock`: drops `@typescript/native-preview` and its 7 platform binaries.

## web/next: typescript 6 via a named catalog

`next build` loads the TypeScript **JS API** for its build-time type check (`verify-typescript-setup.js` does `require('typescript')` and hands the module to `runTypeCheck`). The native `typescript` 7 package ships only `lib/tsc.js`, `lib/getExePath.js`, and `lib/version.cjs`, no `lib/typescript.js`, so there is nothing to hand over.

The important part: this was already broken on canary, silently. Next detects `@typescript/native-preview` and **skips** the check rather than failing. Forced builds on both sides:

```
canary:     Running TypeScript ...
            Detected @typescript/native-preview as TypeScript compiler. Some Next.js
            TypeScript features (like type checking during build) require the standard
            typescript package.
            Finished TypeScript in 82ms ...     <- skipped

this PR:    Running TypeScript ...
            Finished TypeScript in 5.2s ...     <- actually type-checks
```

Fix: a named catalog entry (`catalogs.next`) pins `typescript` 6.0.3 (the feature-frozen JS twin of 7.0) for `web/next` only, referenced as `catalog:next`. `deps-manager.ts` now preserves a `catalog:<name>` reference instead of normalizing it to `catalog:`; its existing `eachCatalog` helper already walks named groups, so the caret rule and the unused-key report cover them with no extra config. Net effect: Next's build-time type check **runs again**, everything else stays on native TS 7. Retire the named catalog once Next ships TS 7 support (landing in the 16.3 line; 16.2.x is latest stable).

Collapsing `web/next` back to `catalog:` does not fail, it silently turns the check off again, so `tests/github/scripts/deps-manager.test.ts` asserts the reference, the two majors, and the preservation rule. Verified it fails on an injected regression (2 of 6 tests) and passes restored; the `pkg.test.ts` case likewise fails with `catalogs` removed from `MERGE_FIELDS`.

## Verification

Merged current `origin/canary` (the prior run was 19 commits behind, including the batch-writes, paging, and console-waitlist work) and re-verified.

- On the merged base: `format:check`, `lint`, `check-types` (13/13 tasks, **0 cached**, plus the scripts and tests passes under `tsc`), `bun run test` (266/266 across 28 files), and a forced `turbo run build` (7/7, **0 cached**), all green. Also ran the full `bun run clean && bun i && ...` clean-room cycle on the earlier base.
- Confirmed all six dts-emitting packages (`config`, `env`, `auth`, `db`, `cli`, `api/hono`) log `Emit types with typescript@7.0.2`, and that `@web/next:build` ran its TypeScript check as a real cache miss.
- Baselined every `dist` artifact against `canary` in a throwaway worktree: 30 files both sides, and **all 15 `.d.mts` files byte-identical**. The 7 differing `.mjs` files are `packages/env` and the `api/hono` bundle that embeds it, and the only content difference is the baked `__GIT_SHA__`/`__VERSION__` (`280c774a` vs the branch SHA) plus the resulting chunk name. No behavioral change.
- `bun install --frozen-lockfile` passes, and `postinstall` leaves `catalog:next` intact (no `package.json` mutation after a full install).
- Browser pass on the dev stack (web/next now compiling under TS 6): `/`, `/docs`, `/docs/getting-started/scripts`, `/blog`, `/waitlist` all 200 and render with no console errors; API `/api/health` and `/api/openapi.json` 200, Scalar `/api/docs` renders. No UI surface changed, so there is no before/after to compare. Skipped the agent sign-in flow deliberately: it writes a session row to the shared production database, and dist byte-identity already proves runtime behavior is unchanged.


nrjdalal:
Triage against current \`canary\` (2026-08-05): **still wanted, not superseded, but the diff has moved under it.**

\`canary\` is already on \`typescript@^7.0.2\` with the \`catalog:next\` TS 6 pin gone (landed via #783), so the version half of this is done. What this PR still removes is real:

- \`@typescript/native-preview\` is still a dependency (`catalog` -> `^7.0.0-dev.20260707.2`), a second compiler beside `typescript@7`
- `tsgo` still drives `check-types:scripts` and `check-types:tests` in the root `package.json`
- `dts: { tsgo: true }` remains in `packages/config/tsdown.ts`, `packages/config/tsdown.config.ts`, `packages/cli/tsdown.config.ts`, and is dead config: `rolldown-plugin-dts` auto-selects native once the installed TypeScript major is >= 7

Status: `CONFLICTING`, untouched since 2026-07-25, 16 files. Given how much moved (the whole TS 7 migration landed separately), redoing it fresh on `canary` is likely cleaner than rebasing this.

Worth knowing before it is picked up: on Next 16.3, `next build` type-checks under TS 7 again, verified by injecting `const x: number = "str"` and watching the build fail with `TS2322`. The old failure mode was silence, so verify that way rather than by reading timings. Recorded in `.github/notes/plans/next-163-adoption.md`.

PR #744, opened 2026-07-24
feat(marketing): design pass on the landing page

nrjdalal:
Design pass on the marketing landing page, driven by a dual-agent critique (design review + deterministic detector, run in isolation) and several rounds of review.

**Scope:** the landing page and its marketing components only. The form, auth-dialog and `site.legal` work that started on this branch now lives on #746.

## Direction

The direction round rolled two candidate visual worlds and a fused challenger; the maintainer took the standing **canon** exit: the category standard, played straight, at the **Linear + Vercel** craft bar, with **authored abstract UI** instead of literal screenshots. So the canonical structure is preserved and the execution is raised.

## What changed

- **The install command is copyable** and sits directly under the subhead. It was an uncopyable picture while the primary CTA pointed at docs.
- **An authored abstraction of the app shell**, showing only routes that exist (`/dashboard`, `/console`, `/console/docs`) with their real gating. The page sold a starter kit while showing no interface at all. An earlier revision of this invented five routes and claimed in a comment that they were real; that was corrected.
- **Removes** the `size-32` ghost-icon watermark behind every bento card, the fake window traffic lights, and redundant `role="link"` on four real anchors.
- **The tagline holds hyphenated compounds** with `sm:whitespace-nowrap` and real hyphens, so find-in-page and copy-paste work. The previous U+2011 held the line but broke both. Gated at `sm` because `Production-ready` at `text-4xl` measures 304px against 288px of usable width on a 320px viewport.
- **Copy button** on code blocks, with a failed clipboard write surfaced by toast, matching `CopyAsMarkdown`.

## Accessibility

| Fix | Before | After |
|---|---|---|
| Footer landmark | `footer=0` sitewide (credit bar was a `<div>` in `<main>`) | real `<footer>` sibling of `<main>` |
| Section names | announced "Hero", "What's wired", "Call to action" | `aria-labelledby` on each section's own heading |
| Reduced motion | marquee and pulse ran regardless | keyframes behind `prefers-reduced-motion: no-preference`, `motion-safe:` pulse |
| Marquee pause | none | hover and `focus-within`, on a focusable strip. The Tailwind `group-hover:[animation-play-state:paused]` utility was never generated, so no pause had ever applied |
| Code blocks | focusable but could not scroll their overflow | reachable scroll regions with accessible names |
| Marquee icons | `alt={tech.name}` beside a visible label, announced twice | `alt=""`, verified 0 of 80 |

## Conventions recorded

`.agents/skills/design/SKILL.md` gains a **Motion** section (reduced-motion gating; a pause affordance for anything moving over five seconds) plus landmark rules for `contentinfo` and `aria-labelledby`, and a scroll-region rule. These bind future work beyond this PR.

The Motion section reverses one critique line: it said to move the marquee keyframes into `globals.css`, which the maintainer rejected, so the rule lives beside the section and the skill codifies that marketing keyframes stay out of `globals.css`.

## Not in scope here

Skip link (needs all eight `<main>` landmarks plus the shadcn customize script), the blog index `<h3>`/link-affordance work, `--success` light-mode contrast (2.87:1 status pill), light-mode elevation collapse.

**The duplicate `<h1>` in the auth dialog** and every other form/dialog fix moved to #746, so this branch does not address them.

**RSS** was built here and then removed at the maintainer's request. The parked decision moved to its own docs PR.

## Corrections worth flagging

The critique's top finding, "every Button renders no focus ring", was a **false positive**: the detector agent measured after a programmatic `.focus()`, which never arms `:focus-visible`. A real keyboard Tab shows a 3px ring. No change was made on that premise.

## Verification

Types, lint and the CLI suite green. Landing driven in a browser: 3-line headline, 5 green dots, 2 copy buttons, one `footer` landmark, marquee pauses on focus, zero horizontal overflow.


web/next/src/app/globals.css
```diff
@@ -146,3 +146,19 @@
     @apply mt-0;
   }
 }
+
+/* marketing marquee: animates only when motion is welcome, and pauses on hover (WCAG 2.2.2) */
```
nrjdalal: I don't want this

.agents/skills/design/SKILL.md
nrjdalal, line 1: feels like you made the generic skill zerostarter oriented

web/next/src/components/common/access.tsx
```diff
@@ -132,7 +135,12 @@ export function Access({ labelClassName }: { labelClassName?: string }) {
                           placeholder="you@example.com"
                           disabled={loader === "email"}
                         />
-                        {isInvalid && <FieldError errors={field.state.meta.errors} />}
+                        {isInvalid && (
+                          <FieldError
+                            id={`${field.name}-error`}
+                            errors={field.state.meta.errors.slice(0, 1)}
```
nrjdalal: what is .slicr(0, 1)?

web/next/src/components/common/access.tsx
```diff
@@ -90,7 +90,7 @@ export function Access({ labelClassName }: { labelClassName?: string }) {
       <DialogTrigger render={<Button className="w-24" variant="outline" />}>
         <span className={labelClassName}>Login</span>
       </DialogTrigger>
-      <DialogContent className="max-w-md" initialFocus={false}>
```
nrjdalal, line 93: must have added it for a reason?

web/next/src/app/waitlist/page.tsx
```diff
@@ -137,18 +161,22 @@ export default function WaitlistPage() {
...
                       placeholder="you@example.com"
                       className="h-12 px-4 text-center text-base sm:text-left"
                       disabled={joinWaitlist.isPending}
                     />
                     {isInvalid && (
                       <FieldError
+                        id={`${field.name}-error`}
                         className="mt-1 text-center sm:absolute sm:top-full sm:left-0 sm:text-left"
-                        errors={field.state.meta.errors}
+                        errors={field.state.meta.errors.slice(0, 1)}
```
nrjdalal: what is .slice(0, 1)?

web/next/src/app/rss.xml/route.ts
nrjdalal, line 1: you did make it with existing patterns right?

web/next/src/app/(marketing)/page.tsx
```diff
@@ -196,6 +161,20 @@ function Eyebrow({ children }: { children: ReactNode }) {
   )
 }
 
+// Keeps hyphenated compounds ("world-class") whole across a line break using real hyphens, so
+// find-in-page and copy-paste still work; a U+2011 would hold the line but break both. The
+// separating space stays outside the nowrap span, or it becomes non-breaking too and glues the
+// compound to the next word.
```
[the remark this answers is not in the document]
nrjdalal: Fixed in `56dd960a`, though it took two attempts: my first replacement silently no-oped against a mismatched anchor, so the block survived and a later review caught it still there. Worse, by then the text had also gone stale, still describing the pre-`sm:` behaviour. Both files are single-line now.

web/next/src/app/(marketing)/page.tsx
```diff
@@ -196,6 +161,20 @@ function Eyebrow({ children }: { children: ReactNode }) {
...
+function NoBreakCompounds({ text }: { text: string }) {
+  const words = text.split(" ")
+  return words.map((word, index) => (
+    <Fragment key={`${word}-${index}`}>
+      {word.includes("-") ? <span className="whitespace-nowrap">{word}</span> : word}
+      {index < words.length - 1 ? " " : ""}
+    </Fragment>
+  ))
+}
```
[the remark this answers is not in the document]
nrjdalal: Confirmed by measurement, and it was my bug. At the mobile `text-4xl` (36px), `Production-ready` renders **304px** against **288px** of usable width on a 320px viewport, so it overflowed by 16px. It fits from 360px up.

Fixed in `0f21caeb` with `sm:whitespace-nowrap`, so narrow phones wrap naturally and the desktop line break is unchanged. Verified `scrollWidth - clientWidth == 0`.

web/next/src/app/(marketing)/page.tsx
```diff
@@ -233,407 +212,410 @@ docker compose up --build`
...
+                    <Image
+                      src={tech.icon.dark}
+                      alt={tech.name}
+                      fill
+                      sizes="1rem"
+                      className="hidden dark:block"
+                    />
+                  </span>
+                  <span className="text-sm font-medium">{tech.name}</span>
+                </div>
```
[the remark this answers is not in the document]
nrjdalal: Correct, fixed in `0f21caeb`. The visible `<span>` already carries the name, so the icons are decorative and 80 of them were announcing each tech twice.

Worth noting my first attempt at this silently no-oped too (the anchor did not match), so the icons still had their `alt` after the commit that claimed to fix them. Re-verified properly: 0 of 80 marquee icons carry non-empty `alt`.

web/next/src/app/waitlist/page.tsx
```diff
@@ -4,19 +4,24 @@ import { features, site } from "@packages/config/site"
...
 
 const formSchema = z.object({
-  email: z.email({ error: "Please enter a valid email address." }).max(254),
+  // empty and malformed are separate failures, so each names its own problem
+  email: z
+    .string()
+    .min(1, { error: "Enter your email address to join." })
+    .pipe(
+      z.email({ error: "That does not look like an email address. Check for a typo." }).max(254),
+    ),
```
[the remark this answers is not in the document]
nrjdalal: Right to flag, and my first fix was wrong twice over.

I initially added `.trim().toLowerCase()` to the zod schema, which does nothing: TanStack Form hands the **raw** values to `onSubmit`, so the mutation still sent the unnormalized address. Moved it to the submit handler and verified end to end.

Then a later review found the whole thing redundant: `api/hono/src/routers/waitlist.ts:11` already trims in its schema and line 103 inserts `email.toLowerCase()`. So the API owned this the entire time, and my comment claiming it "dedupes with `onConflictDoNothing`, which a stray space or capital would defeat" was simply false.

`591b22a0` drops the lowercase and the false comment; only a `.trim()` remains, purely so the confirmation echoes a clean address.

web/next/src/app/globals.css
nrjdalal: Done. `globals.css` is byte-identical to `canary` now: `git diff origin/canary...HEAD -- web/next/src/app/globals.css` is empty.

The marquee keyframes live with the section that uses them instead, still wrapped in `@media (prefers-reduced-motion: no-preference)`. The rest of that pass (the blue `--primary`/`--ring`, the light-canvas tint, the darkened `--success`) came out entirely.

.agents/skills/design/SKILL.md
nrjdalal: Fair, and rereading them you were right: three of my five additions were narrating this landing page rather than stating a convention.

Dropped the "green dot is the marketing pill's mark" bullet and the `CodeWindow`/`CodeCard` description outright, and compressed the `NoBreakCompounds` paragraph into one general rule about scrollable regions.

What is left is repo-wide: `contentinfo` landmarks, `aria-labelledby` over hand-written `aria-label`, keyboard-reachable scroll regions, and a Motion section. I have since also scoped the footer rule, which overreached by implying a sitewide guarantee when only the marketing landing has footer content.

web/next/src/components/common/access.tsx
nrjdalal: Mine, and it was wrong. Removed in `11f180b3`.

I added it after seeing two messages render concatenated, but `FieldError` (`components/ui/field.tsx:184`) already dedupes:

```ts
const uniqueErrors = [...new Map(errors.map((e) => [e?.message, e])).values()]
if (uniqueErrors?.length == 1) return uniqueErrors[0]?.message
return <ul>…</ul>   // genuinely different errors render as a list
```

So the slice was redundant for the duplicate case it was meant to solve, and in every other case it **hid real errors**: a field with two genuinely different problems would only show one. It also fought a primitive that already handles this, against the skill's defaults-first rule.

web/next/src/components/common/access.tsx
nrjdalal: Yes, and I missed it. It goes back to `chore: sync v3`, where it replaced `tabIndex={-1}`:

```
- <DialogContent className="max-w-md sm:max-w-md" tabIndex={-1}>
+ <DialogContent className="max-w-md sm:max-w-md" initialFocus={false}>
```

Both forms do one job: stop the dialog autofocusing its first field. Removing it outright fixed the focus trap but reintroduced exactly that, and since `DialogContent` renders `{children}` before the close button, the first focusable with magic-link enabled is the **email input**, which pops the keyboard on mobile. I did not catch it because this environment reports no magic-link provider, so my test saw focus land on a provider button.

`11f180b3` points `initialFocus` at the dialog heading instead. Focus still enters the dialog, so the trap and the screen-reader announcement both work, but no field is autofocused. Verified: focus lands on the `<h2>`, `activeIsInput: false`, and Tab stays trapped.

web/next/src/app/waitlist/page.tsx
nrjdalal: Same as the `access.tsx` one, and same answer: it was mine and it was wrong. `FieldError` already dedupes by message and renders a list only for genuinely different errors, so the slice was redundant at best and hid real errors at worst. Removed in `11f180b3`.

web/next/src/app/rss.xml/route.ts
nrjdalal: Mostly, yes. It follows `sitemap.ts` and `(llms.txt)/llms-full.txt/route.ts`: `force-static`, `revalidate = 60`, module-scope `contentSource("blog")`, `getPublishedBlogPosts()` for the publish gate (same seam the sitemap uses, so a draft cannot leak), `site` from config, and a plain `new Response()`.

Two things I did get wrong and have since fixed:

- I wrote `content-type` lowercase where the other handlers use `Content-Type`.
- RSS 2.0 requires an email address in `<author>`, so shipping a display name there was invalid. It is `dc:creator` now, with the namespace declared.

I deliberately did **not** reuse `llmTextHeaders`; that constant is llms-specific (`text/markdown`), not a generic helper.

web/next/src/app/rss.xml/route.ts
```diff
@@ -0,0 +1,66 @@
...
+  const items = posts
+    .map((page) => {
+      const published = toBlogDate(page.data.publishedAt)
+      const description = page.data.description ?? ""
+      return [
+        "    <item>",
+        `      <title>${escapeXml(page.data.title)}</title>`,
+        `      <link>${baseUrl}${page.url}</link>`,
+        `      <guid isPermaLink="true">${baseUrl}${page.url}</guid>`,
+        `      <pubDate>${published.toUTCString()}</pubDate>`,
```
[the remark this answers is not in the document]
nrjdalal: Correct, fixed in `591b22a0`. I was escaping titles and descriptions but interpolating `baseUrl` and `page.url` raw into `<link>`, `<guid>`, and `<atom:link>`, so a slug or base URL carrying an ampersand would have broken the document. All URLs go through `escapeXml` now.

Same commit also replaces the invalid `<author>` display name with `dc:creator`. Re-verified: 200, `application/rss+xml`, parses, 3 items, zero `<author>` elements.

web/next/src/components/common/access.tsx
```diff
@@ -210,9 +219,29 @@ export function Access({ labelClassName }: { labelClassName?: string }) {
...
+                    href={site.legal.privacy}
+                    target="_blank"
+                    rel="noopener noreferrer"
+                    className="hover:text-foreground underline underline-offset-4"
+                  >
+                    Privacy Policy
+                  </a>
+                  .
+                </div>
+              )}
```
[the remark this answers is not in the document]
nrjdalal: Good catch, fixed in `591b22a0`. The consent block sat inside the `hasAlternatives` branch, so a fork with only magic-link enabled would have set `site.legal` and still seen nothing. It is a sibling of that block now, so it renders whenever both legal URLs are configured, regardless of which providers are on.

web/next/src/app/waitlist/page.tsx
```diff
@@ -4,19 +4,25 @@ import { features, site } from "@packages/config/site"
...
 import { Field, FieldError, FieldLabel } from "@/components/ui/field"
 import { Input } from "@/components/ui/input"
 import { Spinner } from "@/components/ui/spinner"
 import { apiClient, unwrap } from "@/lib/api/client"
 
 const formSchema = z.object({
-  email: z.email({ error: "Please enter a valid email address." }).max(254),
+  // empty and malformed are separate failures, so each names its own problem
+  email: z
+    .string()
```
nrjdalal: why do we need string and email together, see zod 4 docs

web/next/src/components/marketing/code-window.tsx
```diff
@@ -0,0 +1,59 @@
...
+import { RiCheckLine, RiFileCopyLine } from "@remixicon/react"
+import { useEffect, useState } from "react"
+
+import { ShikiRegion } from "@/components/marketing/shiki-region"
+import { Button } from "@/components/ui/button"
+
+function CopyButton({ code, label }: { code: string; label: string }) {
+  const [copied, setCopied] = useState(false)
+
+  useEffect(() => {
```
nrjdalal, line 13: this is implemented in same way one existing other copy button exists?

web/next/src/app/waitlist/page.tsx
nrjdalal: Tested rather than assumed, and you are half right.

`.min(1)` **is** removable: a function `error` on `z.email()` can distinguish the two failures, so the separate min is gone. But `z.string()` is load-bearing for `.trim()`. Measured across four inputs on zod 4.4.3:

| schema | `""` | `"not-an-email"` | `"  Me@Example.COM  "` |
|---|---|---|---|
| `z.string().trim().pipe(z.email({error: fn}))` | Required | Invalid | **OK** |
| `z.email({error: fn})` alone | Required | Invalid | **Invalid** |
| `z.email({error: fn}).trim()` | Required | Invalid | **Invalid** |

`z.email()` runs its format check before any trim, so a padded address is rejected outright. Committed form is now:

```ts
email: z.string().trim().pipe(z.email({ error: (issue) => … }).max(254))
```

Simplifying this also surfaced a real bug it had been masking, see the separate note: the form registered the same schema on `onSubmit`, `onChange` and `onBlur`, and TanStack keeps one error per source, so a corrected field still carried the stale submit error. That only looked fine while both messages were identical and the dedupe collapsed them. One `onChange` validator now. Verified: empty submit blocks with 0 POSTs and focuses the field, malformed shows its own message alone, and a corrected value clears.

web/next/src/components/marketing/code-window.tsx
nrjdalal: Partly, and the gap was worth closing. Compared against `components/docs/copy-as-markdown.tsx`:

Same: `RiCheckLine`/`RiFileCopyLine`, `Button variant="ghost"`, `navigator.clipboard.writeText`, a 2s reset.

Differed:
- **It surfaced a failed write with `toast.error`; mine swallowed it.** Fixed, so a blocked clipboard now says so instead of silently doing nothing.
- Reset timing: mine uses `useEffect` with cleanup rather than a bare `setTimeout` in the handler, so it cannot fire after unmount.
- Affordance: `CopyAsMarkdown` is icon-only inside a `Tooltip` for a dense docs header; this one carries a visible "Copy"/"Copied" label on a code block, where the target is large and the action should be obvious without hover. Deliberate, not drift.

PR #745, opened 2026-07-25
feat(deps-manager): enforce caret ranges for catalog entries

nrjdalal:
## What

Five catalog entries were exact-pinned with no rationale recorded anywhere (`changelogen`, `fumadocs-core`, `fumadocs-ui`, `react-resizable-panels`, `recharts`), and nothing distinguished a deliberate pin from one that arrived by accident. This adds a rule to `deps-manager.ts` that keeps every catalog entry on a caret range, and carets those five in the same diff.

Ported from a downstream fork where it has been running; the script was byte-identical to this one before the change, so this is a clean port rather than a re-derivation.

## Behaviour

| Spec | Action |
| --- | --- |
| `^1.2.3` | left alone |
| `1.2.3`, `4.7`, `5` | rewritten to a caret, reported |
| `~1.2.3` | rewritten to `^1.2.3`, reported |
| `2.0.14-beta.1` | rewritten to `^2.0.14-beta.1`, reported |
| `>=4.4.3 <5`, `2.x`, `latest` | left alone, **reported for a manual fix** |
| `workspace:`, `file:`, `link:`, `portal:`, `npm:` aliases, `github:owner/repo`, git/http URLs | skipped silently |

Rather than mangling a range a caret cannot express, the script leaves it and prints it under `[INFO] Catalog ranges that are not caret and cannot be converted safely`. Rewrites are reported too, so a pin that gets reverted is visible in the install output rather than silent.

Normalization runs **after** the auto-move step, so a spec promoted into the catalog from a workspace dep is caught in the same run. That is the `bun add` path, and it was the one case an earlier revision of this change missed. Covers `catalog` and every group in `catalogs`.

The catalog walk that `collectCatalogKeys` and the new code both needed is now one `eachCatalog` helper.

## Why the five pins are careted here

The rule rewrites them on whoever's next install either way. Doing it in this diff makes the effect reviewable instead of landing silently on someone else's machine. If any of those five were pinned for a reason, this PR is the place to say so, and the pin should come back with that reason recorded in `.github/notes/dependencies.md`.

## Lockfile

Rewriting a spec during `postinstall` happens after resolution, so `bun.lock` keeps the old spec string until the next install. That is why the lockfile in this branch took two installs to settle.

It is safe: `bun install --frozen-lockfile` compares the resolved package set rather than the recorded spec, verified to pass in both directions (widening and narrowing) without rewriting the lockfile. This matters because `auto-labeler.yml` installs with `--frozen-lockfile`. The existing `catalog:` rewrites already had this property, so it is not new behaviour.

## No opt-out, deliberately

The rule reverts an unexplained pin on every install. A considered pin therefore means changing the rule itself (for example an allowlist it skips) and recording why in `.github/notes/dependencies.md`, which both that note and `AGENTS.md` now state. An allowlist was left out as speculative: there is nothing to exempt today.

Enforcement is partial by design. Unconvertible specs are reported and the script still exits 0, so a compound range survives an install with only a line of `postinstall` output. Failing the install would block work for a range that may be perfectly deliberate.

## Verification

- `check-types`, `lint`, `format:check` and `build` all exit 0 in this repo.
- Rule exercised against 27 spec shapes, including both `catalogs` groups, `*`, `v1.2.3`, whitespace, empty string, `>=1.0.0 || ^2.0.0`, `1.0.0 - 2.0.0` and build metadata. All classified as the table above.
- Auto-move path regression-tested with a fixture workspace declaring `left-pad: 1.3.0` and `chalk: ~5.0.1`; both land in the catalog as carets.
- Idempotent: after the two settling installs, a further `bun install` produces no rule output and leaves `package.json` and `bun.lock` untouched.
- `skills-manager.ts` run to confirm it regenerates only the AGENTS.md skills tables and preserves the new Code-section rule.

At the point that section was written the rule commit changed no resolved version: every entry in `bun.lock` matched `canary` and only the five spec strings differed. Two later commits deliberately change that, below.

## Lockfile refreshed so the carets mean something

A caret that never resolves forward is an exact pin with extra characters, so `bun.lock` was deleted and rebuilt. **175 packages move, 3 arrive, 1 leaves.** The two entries this branch careted that were not already at their ceiling move with it: `fumadocs-core` and `fumadocs-ui` 16.11.5 to 16.12.1, and `recharts` 3.9.2 to 3.10.0. `changelogen` and `react-resizable-panels` were already latest. Framework-level: `react` 19.2.7 to 19.2.8, `hono` 4.12.31 to 4.12.32, `better-auth` 1.6.23 to 1.6.25, `oxlint` 1.74 to 1.75. `next`, `tailwindcss`, `typescript` and `drizzle-*` are unchanged.

**This branch was failing `bun audit` before that refresh.** `brace-expansion <=5.0.7` ([GHSA-mh99-v99m-4gvg](https://github.com/advisories/GHSA-mh99-v99m-4gvg), high) reached us through `ts-morph` → `minimatch` with no override covering it, so the canary pre-push gate was red. The fresh resolve lifts it to 5.0.8 and clears it.

## Three of five audit overrides retired

Each was removed, then the tree was re-resolved and re-audited to see what actually came back.

| Override | Outcome |
| --- | --- |
| `fast-uri` | **Retired.** `ajv` and `shadcn` now resolve `3.1.4` unaided, meeting the recorded exit criterion. |
| `shell-quote` | **Retired.** `concurrently` now resolves `1.9.0`, past the `<=1.8.4` range. Exit criterion met. |
| `esbuild` | **Retired by decision**, see below. |
| `postcss` | **Kept.** Still load-bearing: without it `next` and `shadcn` pull `8.4.31` straight back. |
| `sharp` | **Kept.** Still load-bearing: without it `next` pulls `0.34.5` back. |

The `esbuild` override is the one worth reviewing. It existed for [GHSA-gv7w-rqvm-qjhr](https://github.com/advisories/GHSA-gv7w-rqvm-qjhr), **withdrawn on 2026-06-17** for naming the wrong package, and removing it confirms `bun audit` no longer reports that advisory at any level. But removal is not free: `drizzle-kit@0.31.10` still reaches the legacy `@esbuild-kit/esm-loader` chain pinning `esbuild ~0.18.20`, which is in range for [GHSA-67mh-4wv8-2f99](https://github.com/advisories/GHSA-67mh-4wv8-2f99) (**moderate**), where esbuild's dev server answers any website's cross-origin request. Nothing here starts that server, `drizzle-kit` only uses the loader to read `drizzle.config.ts`, so it is recorded under a new **Retired overrides** heading as an accepted exposure rather than a resolved one, with the condition to reinstate it.

Net: `bun audit --audit-level high` exits 0 (it did not before), and the full audit reports two moderates, the `esbuild` one above and the pre-existing `@hono/node-server` via `shadcn`.

## Verification of the refresh

`turbo run build test --force` at **9 of 9 tasks, 0 cached**, 129 tests passing. `check-types` 13/13 with every esbuild-touching task a genuine cache miss (confirmed from the turbo run summaries, not just green output), `lint` and `format:check` clean, and the pre-commit hook re-ran the full build on each commit. Driven in a real browser against the worktree dev server: landing, `/docs`, a docs page, docs search on the upgraded fumadocs, `/blog`, a blog post and `/waitlist`, all 200 with zero console errors.

https://claude.ai/code/session_014PyRGtfRXYzpoPhEoqRR8T

## Summary by CodeRabbit

* **Improvements**
  * Dependency version handling now consistently uses flexible caret ranges where supported.
  * Installation checks now identify and report dependency specifications that require manual review.
  * Dependency override settings were updated to reflect current security and compatibility requirements.

* **Documentation**
  * Updated contributor guidance and audit records to clarify version-range enforcement and retired dependency overrides.


nrjdalal:
### CI note: the audit failure here is pre-existing, not from this change

`Audit, Lint, and Build` is red on `brace-expansion <=5.0.7` ([GHSA-mh99-v99m-4gvg](https://github.com/advisories/GHSA-mh99-v99m-4gvg)). That is inherited from `canary`, not introduced here:

- `canary`'s committed `bun.lock` already resolves `brace-expansion@5.0.7`.
- This branch's lockfile holds the same version, and the commit does not touch that entry.
- The last green run on `canary` predates it; the run on `ced7194b` is `action_required` and never executed, so the failure was latent rather than visible.

**No override is needed.** `minimatch@10.2.5` asks for `^5.0.5`, which already admits the patched `5.0.8`. The advisory only surfaces because the committed lockfile pins the older copy and Bun keeps a locked version that still satisfies its range rather than re-resolving. Deleting `bun.lock` and reinstalling picks `5.0.8` unaided, and `bun audit --audit-level high` then exits 0 (verified locally).

That fix costs ~418 lines of lockfile churn, since a full regeneration also deduplicates nested entries. Deliberately kept out of this PR so the caret-rule diff stays reviewable. It wants its own PR, which would unblock this one on rebase.

nrjdalal:
### End-to-end verification

The PR body originally asserted "no runtime or UI surface is touched" without testing it. That was reasoning, not evidence, so here is the evidence.

**Nothing resolves differently.** Diffing every resolved entry in `bun.lock` against `canary`:

```
IDENTICAL -> no resolved version changed
```

The only lockfile changes are the five recorded spec strings (`16.11.5` to `^16.11.5`, etc). `node_modules` is byte-identical to `canary`, which is why nothing could move.

**Driven in a real browser anyway** (agent-browser, 1782x972, dev stack in the worktree):

| Surface | Result |
| --- | --- |
| `/` landing | renders, HTTP 200 |
| `/docs` (fumadocs) | full sidebar, TOC, article, anchor links |
| `/docs` search | typed "database", returned real indexed results |
| `/blog` (fumadocs) | all three posts with titles, dates, excerpts |
| `/api/health` | `{"message":"ok","environment":"local"}` |
| console | **zero errors** across the whole session |
| mobile 390x844 | no horizontal overflow on `/docs` |

`/docs` and `/blog` matter most here: they are the fumadocs surfaces, and `fumadocs-core`/`fumadocs-ui` are two of the five entries this PR unpins. Both render and search works.

The other two UI entries do not render at all: `recharts` and `react-resizable-panels` are imported only by `components/ui/chart.tsx` and `components/ui/resizable.tsx`, and nothing imports those.

Local gates in the worktree: `check-types`, `lint`, `format:check`, `build` all exit 0.

PR #746, opened 2026-07-25
fix(web): rebuild the auth and waitlist forms on the documented pattern

nrjdalal:
Rebuilds the waitlist and auth-dialog forms on the documented TanStack Form pattern, and fixes the accessibility defects a design critique found in them. Split out of #744, which keeps the landing-page work.

## The form pattern

Two upstream sources describe this and they disagree, so the new `forms` skill reconciles them: [shadcn's guide](https://ui.shadcn.com/docs/forms/tanstack-form) owns the composition, [TanStack's validation guide](https://tanstack.com/form/latest/docs/framework/react/guides/validation) owns the validator wiring.

**One validator source.** Canary registered the same schema on `onSubmit`, `onChange` and `onBlur`. TanStack keeps one error per source and documents that they accumulate: *"Since `field.state.meta.errors` is an array, all the relevant errors at a given time are displayed."* So a corrected field still carried the entry from the value at the last submit and `FieldError` listed both. It was invisible only while every source emitted identical text and the dedupe collapsed them; splitting the empty and malformed messages exposed it.

**`onBlur` is that source.** Measured on this codebase rather than argued:

| source | mid-typing | after blur | corrected, still focused | untouched empty submit |
|---|---|---|---|---|
| `onChange` | error after 1 character | error | clears | caught, 0 requests |
| **`onBlur`** | **quiet** | error | holds until blur or submit | **caught, 0 requests** |
| `onSubmit` | quiet | quiet | quiet | **blocked with no message** |
| `onBlur` + `onChange` | error after 1 char | error | **never clears** | caught |

`onSubmit` alone (shadcn's example) fails silently: the field is never marked touched, so `isInvalid` stays false and the visitor clicks submit and nothing happens. The hybrid is the worst of both. `onBlur` never calls a half-typed address invalid yet still catches a never-focused empty field, because submitting blurs the field first.

## Accessibility

| Fix | Before | After |
|---|---|---|
| Auth dialog focus | `activeInDialog: false`; 4 Tabs walked into `aria-hidden` content | `initialFocus` at the dialog heading: focus enters for the trap and the announcement, without autofocusing the email field |
| Duplicate `<h1>` | 2 with the dialog open | 1 |
| Waitlist success | bare `<div>`, focus fell to `<body>` | `role="status"`, takes focus, echoes the address, offers a way back and a step forward |
| Empty vs malformed | one message for both | each names its own problem |
| Autofill | none | `autoComplete` + `inputMode="email"`, `aria-describedby` → `FieldError id` |
| Failed submit | focus stranded on the button | `onSubmitInvalid` focuses the invalid field |

`initialFocus={false}` was deliberate upstream (it replaced `tabIndex={-1}` in the v3 sync) to stop the dialog autofocusing its first field. Removing it outright fixed the trap but reintroduced that, since `DialogContent` renders children before the close button. Pointing it at the heading satisfies both.

## Consent line

It named a Terms of Service and a Privacy Policy that were plain text, asserting agreement to documents nobody could open. Both now come from `site.legal`, render only when a provider exists to click, and the sentence hides entirely when a fork has not set them.

**Fork safety:** `site.legal` without a matching CLI template would have thrown in every `zerostarter init` fork and failed its `check-types`, since `Site` is `typeof site`. The template emits it, and `templates.test.ts` now parses the top-level keys out of `packages/config/src/site.ts` and diffs them against the template's, rather than checking a hand-written list. Verified by adding a key upstream only: a hand-written list passes green, this fails.

## Skill

`.agents/skills/forms/SKILL.md` records the shape, the reconciliation, the measured validator comparison, the success-state requirements, and a four-point browser checklist, so the next form does not rediscover any of it.

## Verification

Types, lint, and 122 CLI tests green. Forms driven in a browser: quiet while typing, untouched empty submit shows the required message with **no network request** and focuses the field, malformed shows its own message after blur, and correcting an errored value then clicking submit directly clears the error and submits.


nrjdalal:
Triage against current `canary` (2026-08-05): **needs a scope re-check before rebase.**

`@tanstack/react-form` is already used in several places on `canary` (`waitlist/page.tsx`, `components/common/access.tsx`, and console data tables), so parts of "the documented pattern" appear to be in place. Whether the auth forms specifically still diverge is the question this PR should be re-scoped around.

Status: `CONFLICTING`, untouched since 2026-07-25.

One hazard worth flagging on all four stale PRs: a `CONFLICTING` PR runs no `pull_request` workflows, so `auto-check-build` is absent rather than failing and `gh pr checks` can look green. Do not read these as passing until they are rebased onto `canary`.

PR #752, opened 2026-07-25
fix(cli): honor skill provenance on sync, fix the marketing-fonts abort and --outdated drift

### LINKED ISSUE #749: sync aborts on a fork that owns web/next/src/lib/marketing/
`zerostarter@0.1.12`

## What happens

`bunx zerostarter sync` in a fork that has `web/next/src/lib/marketing/fonts.ts` aborts and rolls back:

```
Sync failed; rolled the working tree back to your last commit.

  web/next/src/lib/marketing/fonts.ts survived the fork strip (add web/next/src/lib/marketing/ to .gitpickignore).
```

The fork owns that file legitimately: it declares the landing's marketing-only display face, kept out of `lib/fonts.ts` so its preload stays on the routes that import it. `convert.ts`'s sibling check enforces exactly that placement ("author-only marketing fonts must live in the fork-excluded `web/next/src/lib/marketing/`, not the shared `fonts.ts`"), so a fork following the rule is guaranteed to trip this.

## The suggested fix cannot work from a fork

`.gitpickignore` on `main` and `canary` already lists `web/next/src/lib/marketing/`. And `sync` fetches the ignore file from `raw.githubusercontent.com/nrjdalal/zerostarter/<ref>/.gitpickignore`, not from the working tree, so a local one has no effect. Verified: adding a `.gitpickignore` containing exactly `web/next/src/lib/marketing/` reproduces a byte-identical failure.

## This one check is the whole blocker

Confirmed by experiment. In a throwaway worktree, `git rm web/next/src/lib/marketing/fonts.ts`, commit, then `bunx zerostarter sync`:

```
◆  Review the changes
│  Starter files were updated (edits overwritten); your content, public/marketing, and branding were preserved.
└  Synced to the latest ZeroStarter
```

It completes and lands a 15-file diff. `web/next/src/lib/marketing/` correctly stays absent afterwards (it is gitpick-excluded), so nothing else in the pipeline depends on the file being gone.

## Reading of the cause

From `dist/convert-*.mjs` (minified, so treat as inferred): the throw is an unconditional existence check in the post-strip validator, which reads as an `init`/`reinit` invariant, don't leak the canonical repo's author-only marketing fonts into a new fork, that also runs on the `sync` path, where a fork owning that directory is the expected state.

**Impact:** no fork with a marketing font can re-baseline.

### LINKED ISSUE #750: sync overwrites customized skills and restamps local/vendored provenance as upstream
`zerostarter@0.1.12`

## The documented contract

From `.agents/skills/skills-manager/SKILL.md`:

> Each skill carries its **provenance** in frontmatter:
> - `source: local` is authored here (this repo is the origin).
> - `source: <tool>` is vendored: re-synced by re-running the tool, never hand-edited.
> - `source: https://github.com/<owner>/<repo>` marks a skill a fork syncs from upstream.
>
> That note is the contract: `bunx zerostarter` updates a skill only while the note is intact **and the body still matches upstream**. Customize the skill or drop the note and the fork owns it.

## What sync actually does

It rewrites every skill regardless of provenance. Observed in a fork with 21 skills (this is the sync run from #749, taken after the blocker there was worked around):

| Skill | Before | After sync |
| --- | --- | --- |
| `design` | note intact, body customized | 4 fork-only bullets **deleted**, upstream body restored |
| `codebase-map` | note intact, body customized | fork's route-group row overwritten with upstream's |
| `vendor` | `source: local`, no note | restamped `source: https://github.com/nrjdalal/zerostarter` + note added |
| `agent-browser` | `source: agent-browser` | restamped `source: https://github.com/nrjdalal/zerostarter` + note added |
| `portless` | `source: portless` | restamped `source: https://github.com/nrjdalal/zerostarter` + note added |

Two distinct violations:

**1. Customized bodies are overwritten.** Per the contract, a customized body means the fork owns the skill and sync leaves it. Instead `design` lost its entire brand section, the bullets documenting the fork's brand color and its contrast math, its filled-panel token, its chart ramp, and its landing display face. These are the fork's design language, not starter content, and nothing about the deletion is announced beyond "edits overwritten" in the summary.

**2. Provenance is fabricated on skills that never came from upstream.** `vendor` is fork-authored (`source: local`, no sync note) and comes back claiming to be synced from zerostarter. `agent-browser` and `portless` are tool-vendored and get the same stamp, which also moves them out of the **Vendored** table in the generated `AGENTS.md` and into **Custom**. After one sync, the frontmatter no longer records where any skill actually came from, and the next sync will happily overwrite the fork-authored ones too.

The second is the more damaging half: #1 is recoverable from the diff, but a wrong `source` is a silent, self-propagating loss of provenance.

## Expected

Skip any skill whose body diverges from upstream or whose `source` is not this repo, and leave a non-upstream `source` untouched. If overwriting a customized skill is intended, it needs to be opt-in and named in the summary, not folded into "edits overwritten".

### LINKED ISSUE #751: skills-manager --outdated always reports DIFFERS in a fork
`.github/scripts/skills-manager.ts`

## What happens

In a fork, `bun .github/scripts/skills-manager.ts --outdated` reports `DIFFERS from upstream` for every skill whose `source` points at zerostarter, including skills that are byte-identical to upstream modulo the fork transform. Observed: 18 of 18 upstream-sourced skills reported `DIFFERS` in a fork where only 2 had genuinely drifted.

## Cause

`outdated()` compares the raw body after the frontmatter:

```ts
const body = (t: string) => t.slice(t.indexOf("---", 3) + 3).trim()
const ours = body(await Bun.file(s.file).text())
const theirs = body(await res.text())
console.log(`  ${s.name}: ${ours === theirs ? "up to date" : "DIFFERS from upstream"}`)
```

But the fork transform edits that same body: it inserts the

```
> [!CAUTION]
> Synced from https://github.com/nrjdalal/zerostarter. ...
```

note at the top, and rebrands prose to the fork's project name (`the ZeroStarter dev stack` -> `the <fork> dev stack`, `portless get zerostarter` -> `portless get <fork>`). So the comparison is against a string the sync itself guarantees will not match, and the check can never print `up to date` for a synced skill.

It does not surface in this repo because every skill here is `source: local`, so the branch never runs.

## Impact

The signal is 100% false positives, which makes real drift invisible. In the fork above, two skills had actually fallen behind a skills rewrite; finding them meant normalizing the transform by hand and diffing, because `--outdated` said the same thing about all 18.

## Suggested fix

Stamping the upstream commit (or a hash of the upstream body) into frontmatter at sync time and comparing that is more robust than normalizing prose, since the brand rename touches arbitrary text and a fork name can appear in a code fence, a URL, or a log path. Cheaper alternative if the frontmatter contract should not change: strip the sync note and fold both project names to a placeholder before comparing.

Worth pairing with #750: once `source` is preserved correctly, `--outdated` becomes the check that tells a fork what it is actually missing.

nrjdalal:
Fixes #749, fixes #750, fixes #751, and updates the AGENTS.md worktree rule.

## What changed

**#749 - sync aborted on a fork that owns `web/next/src/lib/marketing/`**
The unconditional existence check in `fixDangling` was an init/reinit invariant (verify the fork strip removed the author-only marketing module) that also ran on the sync path, where a fork owning that directory is the expected state. It now lives in `assertMarketingStripped`, called only from `convertRepo` (init/reinit); `sync` keeps the navbar strip and the `fonts/marketing/` leak check, which stay valid after an overlay.

**#750 - sync overwrote customized skills and restamped local/vendored provenance**
`sync` now snapshots every `.agents/skills/*/SKILL.md` before the overlay and reconciles per skill from that snapshot, honoring the documented contract:

- `source` not the upstream repo (`local`, a tool, another repo): the fork's file is restored byte-for-byte, never restamped.
- Sync note removed, or body customized since the last sync: the fork's file is restored untouched.
- Note intact and body verified unchanged: the fresh upstream body is taken (rebranded, as before).
- Fork-only skills the overlay never touched are left alone.

Verification uses two new frontmatter stamps written on every transform: `source-hash` (the upstream body the skill was synced from) and `synced-hash` (the body as written, which the next sync hashes against to detect customization). A note-intact skill with no stamp (a fork synced before stamps existed) is overwritten once to gain its stamps and is named in the sync summary, so a lost customization is recoverable from the reviewable diff; every category is reported by name instead of being folded into "edits overwritten".

**#751 - `--outdated` reported DIFFERS for every synced skill in a fork**
The raw-body comparison could never match in a fork (the transform inserts the CAUTION note and rebrands prose). When raw bodies differ, `skills-manager.ts` now hashes the fresh upstream body and compares it to the stamped `source-hash`: match reports `up to date`, mismatch reports `OUTDATED (upstream changed since the last sync)`, and a missing stamp reports `unverifiable` instead of a false positive. This is the frontmatter-stamp fix the issue suggested, chosen over prose normalization since a fork name can appear in code fences, URLs, and log paths.

**AGENTS.md** - the worktree rule now prescribes forking the branch from the latest `canary` (`git fetch origin canary && git worktree add .claude/worktrees/<name> -b <branch> origin/canary`) instead of merging canary in after the fact.

## Notes

- Stamps are double-quoted in frontmatter: strict YAML parses an unquoted all-digit (or `digits+e`) hash as a number, and `skills-manager`'s parser would then drop it. Caught by end-to-end simulation, covered in the skill doc.
- Both hash implementations (`packages/cli/src/skills.ts`, `.github/scripts/skills-manager.ts`) are 16 hex chars of sha256 over the LF-normalized trimmed body and must stay identical; both files say so.
- Docs synced in the same change: `getting-started/cli.mdx` (sync section) and the `skills-manager` skill (contract, stamps, new `--outdated` outputs).

## Verification

- `packages/cli`: 129 tests pass (new coverage: sync tolerates fork-owned marketing fonts; init still asserts the strip; per-outcome sync cases for added/updated/customized/note-removed/non-upstream/fork-only skills).
- `bun run check-types`, `bun run lint`, `skills-manager --check` all clean.
- End-to-end simulation against live upstream (scratch fork + the new `skills-manager.ts`): stamped-and-current reports `up to date`, stale stamp reports `OUTDATED`, missing stamp reports `unverifiable`; the `node:crypto` and `Bun.CryptoHasher` hashes agree byte-for-byte.

## Summary by CodeRabbit

* **New Features**
  * Improved skill synchronization with provenance and hash tracking.
  * Preserves customized or unverifiable skills while updating unchanged upstream skills.
  * Sync results now summarize added, updated, kept, skipped, and unverifiable skills.

* **Bug Fixes**
  * Prevents customized fork skills from being unintentionally overwritten during sync.
  * Improves handling of missing or malformed synchronization metadata.

* **Documentation**
  * Updated CLI guidance to explain skill preservation and synchronization behavior.

### DIFF
--- a/.agents/skills/skills-manager/SKILL.md
+++ b/.agents/skills/skills-manager/SKILL.md
@@ -14,7 +14,7 @@ Each skill carries its **provenance** in frontmatter:
 
 - `source: local` is authored here (this repo is the origin).
 - `source: <tool>` is vendored: re-synced by re-running the tool, never hand-edited.
-- `source: https://github.com/<owner>/<repo>` marks a skill a fork syncs from upstream (the CLI stamps this, with a `[!CAUTION]` sync note, when it scaffolds a fork).
+- `source: https://github.com/<owner>/<repo>` marks a skill a fork syncs from upstream (the CLI stamps this, with a `[!CAUTION]` sync note, when it scaffolds a fork). A synced skill also carries two drift stamps: `source-hash` (the upstream body it was last synced from) and `synced-hash` (the body as written by that sync). Both are 16 hex chars of sha256 over the LF-normalized, trimmed body, always double-quoted (an all-digit hash would otherwise parse as a YAML number and vanish); the implementations in `packages/cli/src/skills.ts` and `.github/scripts/skills-manager.ts` must stay identical.
 
 ## Regenerate the tables
 
@@ -29,13 +29,13 @@ Each cell is the description's summary, the sentence before `Use ...`, so keep e
 
 ## How a fork syncs
 
-A fork inherits its skills from the scaffold. On `init` and `sync` the CLI rebrands each skill's prose to the fork's project name (read from `package.json`) and marks it with `source: <upstream repo>` plus a `[!CAUTION]` note at the top. That note is the contract: `bunx zerostarter` updates a skill only while the note is intact and the body still matches upstream. Customize the skill or drop the note and the fork owns it. Check state with:
+A fork inherits its skills from the scaffold. On `init` the CLI rebrands each skill's prose to the fork's project name (read from `package.json`) and marks it with `source: <upstream repo>`, a `[!CAUTION]` note at the top, and the two drift stamps. That is the contract, and `sync` enforces it per skill from a pre-overlay snapshot: it takes the fresh upstream body only when `source` is the upstream repo, the note is intact, and the body still hashes to `synced-hash` (nothing was customized since the last sync). A non-upstream `source` (local, a tool, another repo), a removed note, or a customized body leaves the fork's file exactly as it was. The one exception is a note-intact skill with no stamp (synced before stamps existed): it is overwritten once to gain its stamps, and named in the sync summary so a lost customization is recoverable from the diff. Check state with:
 
 ```bash
 bun .github/scripts/skills-manager.ts --outdated
 ```
 
-which reports each skill as `local, no upstream`, `vendored (<tool>)`, or, for a synced skill, `up to date` / `DIFFERS from upstream`.
+which reports each skill as `local, no upstream`, `vendored (<tool>)`, or, for a synced skill, `up to date` / `OUTDATED (upstream changed since the last sync)`, comparing `source-hash` against a fresh upstream fetch (the sync transform guarantees raw bodies never match in a fork), and `unverifiable` when the stamp is missing.
 
 ## Notes
 
--- a/.github/scripts/skills-manager.ts
+++ b/.github/scripts/skills-manager.ts
@@ -7,6 +7,7 @@ import { $, Glob } from "bun"
 //   bun .github/scripts/skills-manager.ts --check    fail on drift instead of writing (the pre-commit gate)
 //   bun .github/scripts/skills-manager.ts --outdated fetch each skill's source upstream and report drift
 // A skill's source is its provenance: a full github link it was inherited from (checked by --outdated), a bare tool name (vendored, re-run the tool), or local (authored here).
+// A synced skill also carries source-hash, the hash of the upstream body it was last synced from (stamped by bunx zerostarter); --outdated compares that against a fresh upstream fetch, since the sync transform (rebrand + sync note) guarantees the raw bodies never match in a fork.
 
 const ROOT = path.resolve(import.meta.dir, "../..")
 const SKILLS_DIR = path.join(ROOT, ".agents/skills")
@@ -18,6 +19,7 @@ type Skill = {
   description: string
   summary: string
   source: string
+  sourceHash: string
   dir: string
   file: string
 }
@@ -62,6 +64,7 @@ async function readSkills(): Promise<Skill[]> {
       description: fm.description,
       summary: summaryOf(fm.description),
       source: fm.source,
+      sourceHash: fm["source-hash"] ?? "",
       dir: rel.split(/[/\\]/)[0]!,
       file,
     })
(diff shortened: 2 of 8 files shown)


nrjdalal:
Superseded by #777, now merged as `b2978b88`.

This PR reached the same conclusion first, that prose cannot be compared because sync rebrands the body, so the fix has to be a recorded hash. #777 carries that idea through, storing the hashes in `.agents/skills/.sync.json` rather than in frontmatter, which sidesteps the all-digit-hash YAML quoting problem this PR had to document.

Also carried over verbatim: the AGENTS.md worktree rule (fork the branch from `origin/canary` rather than merging canary in afterwards).

Closing rather than rebasing because this branch has gone stale, its tests live in `packages/cli/test/`, a per-package location the repo no longer uses, and it conflicts with current canary. The three issues it fixed (#749, #750, #751) are closed by #777.

PR #753, opened 2026-07-25
feat: platform data table with virtualized infinite scroll and a console users page

nrjdalal:
## Summary

Adds the platform data-table: a composable component family over TanStack Table (sourced from the shadcn data-table guide and TanStack's own examples, ported to this repo's base-ui + remixicon dialect), URL-synced state via nuqs, **virtualized infinite scroll** per TanStack Table's virtualized-infinite-scrolling example (react-table + react-virtual + `useInfiniteQuery`), a new admin users API, and the console's Platform > Users page as its first real consumer. Follows the shadcn stance: no all-in-one prop machine; the page owns the table instance and composes the pieces.

## What's inside

- **`components/data-table/`**: `DataTable` (a focusable, viewport-filling region: sticky header, virtualized rows windowed by TanStack Virtual, load-more within 500px of the bottom, initial spinner, `Empty` fallback, inline retry on post-load failures, selection/total status line, explicit table roles + `aria-sort`/`aria-rowcount`), `DataTableCellText` (text-cell overflow: truncate with a tooltip only when actually cut, or wrap with the measured row growing), `DataTableColumnHeader`, `DataTableToolbar` (global-filter search + children slot), `DataTableViewOptions` (labels from `columnDef.meta.label`), `DataTableFacetedFilter`. Layout lives in the column manager (`COLUMN_MANAGER.<area>.<table>.<columnId>` holding `{ align?, flex?, width? }` in Tailwind spacing units rendered via `calc(var(--spacing) * n)`, folded into the defs by `applyColumnManager`/useDataTable's `manager` option): exact widths on fixed columns, floors on flex ones (a run of consecutive flex columns collapses to its last member growing, and whatever follows the last flex column docks to the right edge), optional center/right alignment, and widthless columns measure their header label via `@chenglou/pretext` (canvas metrics, applied post-mount behind a stable fallback). Columns files carry content only; content never shifts the layout. No numbered pagination anywhere.
- **`hooks/use-data-table.ts`**: the whole server-driven wiring in one generic call: URL state, the infinite query (`getNextPageParam` against the response total, `keepPreviousData`, deferred search), and the manual-mode table instance. A consumer brings only what a generic hook cannot know: columns, a typed `fetchPage`, filter ids.
- **`hooks/use-data-table-state.ts`**: the URL layer alone (`q`, `sort` as `column.asc|desc`, one array param per filter id via nuqs), for client-side tables. No page state: a state change resets the list.
- **`GET /api/v1/admin/users`**: server-driven list endpoint (drizzle-direct): page/perPage batches, whitelisted sort + dir, escaped `LIKE` search across name OR email, role filter with null-as-"user" handling. Gated by the new `adminMiddleware` (403 `FORBIDDEN`, same rule as the `/console` gate and the same freshness: the session cookie cache is bypassed, so a revoke locks the API on the next request) and documented via `forbiddenErrorResponses`.
- **`/console/users`**: the console gains a grouped sidebar (Getting Started with Documentation, Platform with Users) and the users table lives at its own Platform page, structured shadcn-style: a `(platform)` route group with the page's pieces colocated at `users/components/data-columns.tsx` + `data-table.tsx` (just columns + a typed `fetchPage` + composition). Server sorting/search/role facet, 25-row batches streaming in on scroll, virtualized rows, row selection, row actions, error state with retry. Columns cover what the data has and an admin needs, rendered as plain text for now: name, email, role, a Status column from the admin plugin's banned flag, and joined date; id stays behind the copy action, and image/emailVerified remain in the payload for a later visual treatment. The layout 404s non-admins and the API gates every row, so the page stays synchronous: the shell paints instantly on navigation (~150ms measured, no route-spinner phase) and only the table's data spinner shows. The page opts into the app-frame height chain (`h-svh` on `PageShell`, `flex-1 min-h-0` down to the region) so the table takes all available height and never resizes with its content; the console landing returns to a minimal home.
- **Docs + skills**: new `manage/data-tables` page (pieces, hooks, virtualized infinite scroll, the viewport-filling chain, server/client recipes); architecture/project-structure/authentication/api-conventions updated where they described the console as empty; `design`, `api-endpoint`, and `codebase-map` skills synced.

## The React Compiler gotcha (worth knowing)

TanStack Table v8 mutates one stable table instance during render, so with `reactCompiler: true` every memoized consumer comparing the instance identity froze one render behind (sort clicks updated the URL but not the rows). Every file reading a `table`/`column` instance now carries `"use no memo"`, per TanStack's v8 guidance, and the docs call it out so new consumers keep it.

Dashboard members table was built and then dropped per review direction (console-only for now); the family supports client-side tables and the docs include that recipe.

No screenshots attached: the console renders real user records, so verification is manual on a local run.

## Verification

- `curl` matrix on the endpoint: 200 as admin with q/sort/dir/role/paging, 401 unauthenticated, 403 as non-admin on the very next request after a revoke (verified against a cache-intact cookie jar), 400 with zod issues on bad `sort`/`role`; OpenAPI lists the route with 200/400/401/403/429/500.
- agent-browser E2E on `/console/users` as LocalAgent: sidebar groups render with correct active states (exact-matched Dashboard link), region fills the viewport with zero page scroll (measured), virtualization confirmed (window of ~22 rendered rows against a larger loaded set, tbody sized by the virtualizer), first batch only on load, scrolling streams the next batch and the status line updates to N of N, sticky header stays pinned mid-scroll, sort asc/desc resets to a fresh sorted batch scrolled to top (URL + first row verified), search (`q`, deferred), role facet composed with sort + search (`?sort=name.asc&role=admin` verified end to end), selection count, reset (clears filters, keeps sort), column hide/restore via View, focusable scroll region, light + dark themes. 375px pass: fixed columns hold their size and the region scrolls horizontally (924px content in a 341px region) instead of crushing cells. Explicit table roles survive the grid/flex flip (5 columnheaders in the a11y tree, `aria-rowcount`/`aria-rowindex` reflect the full virtualized set).
- `bun run check-types` green across all workspaces; oxlint + oxfmt clean.

https://claude.ai/code/session_01RpGEatn22yd6GfW7VCbnUo

## Summary by CodeRabbit

* **New Features**
  * Added an admin-only Users page with searchable, sortable, filterable, virtualized tables and infinite scrolling.
  * Added user status, role, join date, selection, column visibility, and copy-to-clipboard actions.
  * Added URL-synchronized table filters and sorting.
  * Added admin access enforcement with clear unauthorized and forbidden responses.

* **Documentation**
  * Added data-table guidance and updated console, authentication, architecture, and API documentation.

### DIFF
--- a/.github/notes/plans/README.md
+++ b/.github/notes/plans/README.md
@@ -13,6 +13,7 @@ This is the internal, fork-excluded backlog. It is separate from the published `
 
 ## Planned
 
+- [Platform data table](data-table.md) - full PRD; working spike on `spike/platform-data-table`, rebuild clean from the spec (PR #753).
 - [Dynamic per-branch preview URLs](dynamic-preview-urls.md) - predictable per-branch preview URLs on our own domains via alias-on-deploy (#677).
 
 ## Backlog / ideas
--- a/.github/notes/plans/data-table.md
+++ b/.github/notes/plans/data-table.md
@@ -1,0 +1,120 @@
+# Platform data table (PRD)
+
+- Status: planned
+- Links: PR #753 (this plan); working spike on `spike/platform-data-table` (~40 commits of iteration, all requirements below verified live there); shadcn data-table guide; TanStack Table v8 virtualized-infinite-scrolling example
+
+One data table for the whole platform, rebuilt clean from this spec. The spike proved every requirement end to end but its history is iterative churn; the rebuild implements this document directly and ports code from the spike where it is already right. Every requirement below was converged on deliberately, most after seeing the alternative fail.
+
+## Stance
+
+- The shadcn stance, held throughout: no all-in-one `<DataTable data={...} />` prop machine. The page owns its `useReactTable` instance; the family provides composable pieces.
+- Follow documented upstream patterns, never hand-rolled ones: TanStack Table v8 (headless), TanStack Query `useInfiniteQuery` (batches), TanStack Virtual (windowed rows, per their official virtualized-infinite-scrolling example), nuqs (URL state), `@chenglou/pretext` (text measurement).
+- Every table infinite-scrolls inside its own region. No numbered pagination anywhere.
+- Cells render plain text for now: no badges, avatars, or icons in data cells (the payload keeps `image`/`emailVerified` for a later visual treatment).
+
+## Architecture
+
+```
+components/data-table/            the family: generic rendering, no data knowledge
+  data-table.tsx                  DataTable (namesake file matches the export)
+  cell-text.tsx                   DataTableCellText: truncate | wrap overflow modes
+  column-header.tsx               DataTableColumnHeader: title + bare sort icon
+  column-manager.ts               COLUMN_MANAGER + applyColumnManager (all layout)
+  faceted-filter.tsx              DataTableFacetedFilter (Popover + Command)
+  toolbar.tsx                     DataTableToolbar (search + children slot + View)
+  view-options.tsx                DataTableViewOptions (labels from meta.label)
+hooks/
+  use-data-table.ts               server wiring: URL state + infinite query + table
+  use-data-table-state.ts         URL layer alone (client-side tables)
+app/(console)/console/(platform)/users/
+  page.tsx                        synchronous server component (no page gate)
+  components/data-columns.tsx     content only: header, cell, meta.label
+  components/data-table.tsx       UsersDataTable: fetchPage + composition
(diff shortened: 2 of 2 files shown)


nrjdalal:
Superseded by the plan-first flow: the full PRD is up in #754 and the rebuild will follow it in a clean context. The working implementation from this PR is preserved on `spike/platform-data-table` (every PRD requirement was verified live here); the review findings in this thread are folded into the PRD's requirements and traps sections.

PR #797, opened 2026-08-11
refactor: deepen the api client envelope and split the admin router

nrjdalal:
Acts on an architecture review of the repo. 11 commits, 39 files, +1432/−1207. No behaviour change to the app; the one change that did alter runtime behaviour was caught in review and reverted, described under Dependencies below.

## What landed

**`refactor(web)`: split `unwrap` out of the api client.** `unwrap` is the whole error contract for the web app and is pure, but it shared a module with the `hc()` singleton, so importing it validated env and no test could reach it. The envelope reader now lives in `web/next/src/lib/api/unwrap.ts` with no runtime imports; `client.ts` re-exports it and stays the entry to import from. **12 new tests.**

Adds `unwrapOrThrow`. Nine call sites undid the "never throws" promise by hand with `throw new Error(error.message)`, which dropped the envelope's `code` and left callers unable to tell a refusal from a network failure; `ApiRequestError` carries it through. All nine now use it. The three tables that still destructure `{ data, error }` keep `unwrap`, because `runBatched` and `foldBatch` want that shape.

**`refactor(web)`: drop `useDataTable`'s unused query returns.** `error`, `isError` and `refetch` were returned but destructured by none of the four tables. (`isError` and `onRetry` are in `tableProps`, which every caller spreads; `query.error` itself is not, but no caller used it.)

**`refactor(api)`: split the admin router by resource.** `admin.ts` was 1057 lines, 54% of the service, and already sectioned by resource at its feature-gate comments. Now one file per resource behind a single `consoleAdminMiddleware` mount, in the original route order. Largest file is `admin/users.ts` at 441 lines.

`deleteSet()` takes the allowlist and waitlist deletes, which differed by eight lines of pure identifier substitution, and owns the transaction, the write-back and the ordering; the caller still writes its own statement so the table keeps its Drizzle types. The part worth testing is `answerDeleted`, which is pure, sits in `batch.ts` beside the rest of the set vocabulary, and has **5 new tests**.

**`refactor(api)`: one `codeSample()` for every documented route.** Started as an admin-only helper; it now lives in `api/hono/src/lib/code-sample.ts` and replaces **all 15** hand-written `x-codeSamples` blocks, including the six in `index.ts`, `v1.ts` and the public `waitlist.ts`. Zero hand-written blocks remain. The cast is `as object`, so it contributes nothing to the route type; I diffed every emitted snippet before and after.

**`docs`:** reconciles the build-script placement rule, which `AGENTS.md` and `add-package/SKILL.md` stated in opposite directions, documents `unwrapOrThrow` and `deleteSet`, and points the docs and skills at the new paths.

## Dependencies

`bun audit --audit-level high` fails on canary as well as this branch (GHSA-2v37-7h3g-55p8, nanoid), so CI could not go green without touching it. Two attempts here were wrong and are recorded in `.github/notes/dependencies.md` under Retired, because that file exists so a returning advisory is not re-investigated from scratch:

- A **bare `"nanoid": "^3.3.17"` override** was wrong. Bun applies a bare override to every copy, and `web/next` depends on nanoid directly (`catalog: ^6.0.1`, used by `customAlphabet` in `src/lib/utils.ts` for `slugify`), so it downgraded a runtime dependency of the shipped app by three majors. Caught in review.
- A **nested `"postcss": { "nanoid": "^3.3.17" }`** was inert. Bun does not record a nested override; `bun.lock`'s map listed `js-yaml` alone, and removing the entry changed no resolution.

What actually cleared it: `postcss@8.5.23` declares `nanoid: "^3.3.16"`, which already admits the patched release, so refreshing the lockfile lifted that one copy. **`overrides` is back to `js-yaml` alone.** The lock carries `3.3.18`, `5.1.16` and `6.0.1`, and audit is clean with no nanoid entry at all.

## Three findings from the review that turned out to be wrong

Each would have been a regression:

- **`features` baked into the auth bundle is not a bug.** Flipping `features.allowlist` cache-misses `@packages/auth:build` and `dist` re-bakes to `allowlist:!1`; restoring flips it back. Verified both directions. Turbo's package graph includes the devDependency, and `packages/auth/tsdown.config.ts:32` documents why it must stay one.
- **The four list endpoints have not drifted.** Waitlist's missing `NULLS LAST` is correct, because both its sortable columns are `notNull`, and the two lists that spell it out are exactly the ones sorting on a nullable expression. Activity has no `sort` param at all. The three count strategies are each right for their query. A unified `listEndpoint` would have flattened correct, documented decisions, so it was not built.
- **Two "dead exports" are live surface.** `notFoundErrorResponses` is documented in `api-endpoint/SKILL.md:58` as the 404 set for by-id routes, and `matchesAllowlist` is named in `plans/signup-gating.md:18` as pre-built for signup gating. Only `useDataTable`'s three returns were actually dead.

## Not in this PR

Written up in `.github/notes/plans/deepening-followups.md` so they survive the merge: the `packages/scripts` build-chain move (needs a Docker build to verify), pointing the package tests at the `exports` map (needs root `@packages/*` devDependencies or a `tests` workspace, and `tests/` is fork-excluded, so it is a design call), the `SessionReader` seam, and deriving the SQL role ladder from `CONSOLE_ROLES`. The last two change auth-shaped structure and want a design pass first.

## Verification

Re-run at head (`7841e390`), after the commits that rewrote every documented route's OpenAPI options and moved the helper:

- `bun run test`: **293 pass**, 0 fail, up from 276 on canary.
- `bun run check-types` and `bun run lint` clean; full `turbo run build` green with real cache misses.
- **RPC surface:** `api/hono/dist/index.d.mts` changes shape, because `.route()` composes a union where the chain was an intersection, but every path keeps its input and output types. `web/next` type-checks against it, and an injected bad sort id still fails through `InferRequestType` (`Type '"bogusColumn"' is not assignable to type 'WaitlistSort'`), so the client types are live rather than degraded to `any`.
- **OpenAPI:** `/api/openapi.json` reports 15 operations carrying `x-codeSamples`, unchanged from before the helper.
- **API, against a disposable pglaunch database seeded with fake rows:** 25 checks, all passing. Every list with paging, search and facets; `sort=role` rank ordering; `NULLS LAST` on `lastActive` and `allowlist.actor`; the 400 validation envelope; both `deleteSet` paths including per-id `NOT_FOUND` with each route's own wording and the activity records they write; the role batch; the guard refusing a self-ban; allowlist create and its 409.
- **`slugify` smoke test** after restoring nanoid 6.0.1: `slugify("Hello World", 6)` gives `hello-world-p3myq0`, and the `size = 0` path returns the bare slug.
- **Browser (ui-verify):** signed in through the agent route, drove all four console tables, ran a two-row removal end to end (confirm dialog to `2 removed` toast to the table going 5 to 3 rows), confirmed both removals in the activity trail by name, and confirmed search narrows to one row with `?q=grace`. No horizontal overflow at 390x844, 834x1112 or 1782x972; both themes render. Details in the ui-verify comment below.

Screenshots are not embedded: litterbox returned 403 for every upload across three rounds from this network. They went to the maintainer directly. Everything in them is seeded fake data (`example.com`) and the local agent account, no real rows.

## Summary by CodeRabbit

* **New Features**
  * Added admin tools for managing users, activity, allowlist rules, and waitlist entries.
  * Added filtering, sorting, pagination, batch actions, authorization safeguards, audit tracking, and feature-gated access.
  * Improved API error handling with consistent typed responses and clearer error details.
* **Documentation**
  * Updated API usage guidance, project structure references, and administrative endpoint documentation.
* **Refactor**
  * Standardized API code samples and response handling across the application.
* **Tests**
  * Expanded coverage for deletion results and API response/error scenarios.


nrjdalal:
## ui-verify pass

Ran against the branch on a disposable pglaunch database seeded with fake rows, portless dev stack, `agent-browser`.

**Stack:** `$WEB/` 200, `$API/api/health` ok.

**Routes driven** (not just loaded) at 1782x972:

| Route | Rows rendered | Horizontal overflow | Error state |
| --- | --- | --- | --- |
| `/console/users` | 13 | none | no |
| `/console/allowlist` | 5 | none | no |
| `/console/waitlist` | 3 | none | no |
| `/console/activity` | 8 | none | no |

**Flow, end to end:** signed in through **Login (agents)**, selected two waitlist rows from the selection bar, confirmed the dialog (`Remove 2 signups?`), and watched the table go 5 to 3 with a `2 removed` toast. Both removals then appeared in `/console/activity` by name (`Removed waiting.alan@example.com from the waitlist`). That path crosses every piece this PR touched: `useDataTable` with its trimmed return, `runBatched`, `unwrap`, the new `deleteSet` on the API, `recordActivity`, `answerDeleted`, `toastBulk`, and the refetch through `unwrapOrThrow`.

Server-driven search narrows the users table to one row with `?q=grace` in the URL, so the fetcher's query building survives the `unwrapOrThrow` change.

**Responsive:** `document.documentElement.scrollWidth <= clientWidth` is `true` at 390x844, 834x1112 and 1782x972.

**Theme:** exercised both. Dark and light each render the table with all 13 rows; light resolves `body` to `lab(100 0 0)`.

**Before/after pairs: consciously N/A.** Nothing in this PR changes styling or markup. The web-side changes are the fetchers' error path (`unwrapOrThrow` instead of `unwrap` plus a hand-written rethrow) and the removal of three `useDataTable` return keys no caller destructured, so a before/after pair would be two identical images. The check that matters here is that the surfaces still render and behave, which the table above records.

**Screenshots not embedded:** litterbox returned 403 for every upload attempt across two rounds from this network, so there are no URLs to embed. Images were handed to the maintainer directly instead. Everything in them is seeded fake data (`example.com`) and the local agent account, no real rows.

nrjdalal:
## Review findings addressed

Every load-bearing claim re-verified against the tree before acting. Three commits: `ceb30eda`, `912a64ea`, `1ecb0ad7`.

### The nanoid override was wrong, and the record was wrong about why

Confirmed all of it. `web/next/package.json` declares `nanoid` as a **direct dependency** (`catalog:` to root `^6.0.1`), `web/next/src/lib/utils.ts:2` imports `customAlphabet` for `slugify`, and the lockfile went from three copies (`3.3.16`, `5.1.16`, `6.0.1`) to a single `3.3.18`. Bun applies a bare override to every copy, so `df82748a` silently downgraded a runtime dependency of the shipped app by three majors inside a PR whose contract is no behaviour change.

Only `postcss`'s `nanoid@3.3.16` was ever in range for the advisory. The override is now scoped under `postcss`, so the lock resolves **3.3.18, 5.1.16 and 6.0.1** and `bun audit --audit-level high` stays clean. Smoke-tested the actual consumer: `slugify("Hello World", 6)` gives `hello-world-p3myq0`, and the `size = 0` path returns the bare slug.

The record is rewritten. All three of its justifications were false, and the real safety argument is the `if (!size) return ""` guard, which it now names. Reordered after js-yaml to match `package.json`.

### codeSample was in the wrong place, and the skill pointed at it

Moved to `api/hono/src/lib/openapi.ts`, beside `error.ts` and `paging.ts`. The six surviving hand-written blocks in `index.ts`, `v1.ts` and `waitlist.ts` now use it too, so **all 15 documented routes** go through one helper and no hand-written block is left. The skill line points at `lib/openapi.ts`. Verified `/api/openapi.json` still reports 15 operations carrying `x-codeSamples`.

### Also fixed

- `type-safe-api.mdx`: the bare `return null` at module top level was invalid. It now branches on the union, which is what the paragraph below it describes.
- `admin/index.ts`: the mount-order comment claimed the feature gates constrain ordering. They sit inside their own routers and constrain nothing; the comment now says so and gives the real reason (stable route list and OpenAPI document).
- `.github/notes/plans/api-envelope-typed-endpoint.md`: still located the envelope shape-sniffing in `client.ts`; it moved to `unwrap.ts`.

### Accepted, not changed

- **`deleteSet` as a thin wrapper.** Fair. It stays because the ordering and the not-found answer are the parts worth having in one place, and `answerDeleted` is the tested half. The `(tx: Transaction)` annotation is load-bearing, not decoration: without it both callbacks are context-sensitive, TypeScript defers both, and `R` falls back to its constraint, which is a real compile error I hit.
- **`query.error` is not in `tableProps`.** Correct, only `errorMessage` and `onRetry` are. The PR body was loose; no caller destructured it either way.

Re-verified after the fixes: 293 tests, lint and types clean, `Audit, Lint, and Build` green, both Vercel previews deployed.

nrjdalal:
## Second review round addressed

`d31f9dd5`, `7841e390`, plus a rewritten PR description.

**N1 — the override was inert, and the record asserted a cause the lock does not support.** Confirmed by experiment, not argument: `bun.lock`'s `overrides` map listed `js-yaml` alone, and removing the nested entry changed no resolution. `postcss@8.5.23` declares `nanoid: "^3.3.16"`, which already admits the patched release, so what cleared the advisory was the lockfile refresh, not anything I added.

`overrides` is back to `js-yaml` alone. The entry moved to **Retired**, with both failed attempts written down and the honest cause named, because that section exists so a returning advisory is recognised rather than re-investigated. Lock still carries `3.3.18`, `5.1.16`, `6.0.1`; audit clean with no nanoid entry.

Worth stating plainly: the review is right that the stated verification could not detect what it claimed. `grep`ing the resolutions proves the tree is fine, not that the override caused it.

**N2** — the skill's "A route that acts on a set" section now documents `deleteSet`, with a worked example and the reason `tx` needs its annotation.

**N3** — `project-structure.mdx:33` now says two files.

**N4** — accepted. `lib/openapi.ts` over-promised for a 5-line file; renamed to `lib/code-sample.ts`, matching `batch.ts`, `paging.ts`, `error.ts`, which are named for what they hold.

**Spec (a)** — PR description rewritten against head: 11 commits, all 15 `codeSample` sites, `admin/users.ts` at 441, both dependency commits described with what went wrong, `codeSample` no longer framed as part of the admin split, and the verification section re-run at `7841e390` rather than carried over.

**Spec (b)** — `plans/deepening-followups.md` added and indexed in `plans/README.md`, covering all four deferred items with why each was left out. `api-envelope-typed-endpoint.md` no longer claims every route hand-writes the `x-codeSamples` block, which `912a64ea` had made false, and links this PR as landing that half.

**Spec (c)** — resolved by the rewrite; the body now matches `admin/index.ts`, which is the correct account.

Re-verified at head: 293 tests, lint and types clean, `Audit, Lint, and Build` green, both Vercel previews deployed, `/api/openapi.json` still reporting 15 operations with `x-codeSamples`.

nrjdalal:
## Third round addressed

`c6160d88`.

**The stale path (both axes, independently).** `api-endpoint/SKILL.md:59` still said `lib/openapi.ts` after the rename to `lib/code-sample.ts`, on the one line the same commit had written two commits earlier, in a commit whose subject is "true up the drifted references". `git grep` confirms it was the tree's only surviving reference and it is now gone. Nothing to say in defence: the rename updated seven importers and a plan file, and I never grepped for the name I had just retired.

**`audit/SKILL.md` was not synced to what this branch did to the file it governs.** Its rule said no block may outlive its override, which forbids the nanoid entry I added, and it never documented the `## Retired overrides` section that entry sits in. It now describes retiring a block (including dropping the `→ <version>` arrow, since there is no version left to record) and carries the lesson that only lived in the notes file: check whether the package is also a direct dependency of a workspace and at which major before reaching for a bare override, because Bun applies one to every copy, and the nested form is not recorded at all. `nanoid` is named as the worked example.

**The ordering contradiction is real and the prose was wrong, not the code.** `deleteSet` passes `targets` through unsorted into one `inArray` statement and answers via `answerDeleted`. I have scoped the sorted-write rule to routes that write row by row, said why a delete set is exempt (one statement takes its locks in one operation), and stopped claiming `deleteSet` owns "the ordering" when what it owns is the per-id answer in the order asked.

**Helper count** no longer says two.

**Two figures in the follow-ups plan.** Both wrong, both corrected: the suite is 293, not 330, and the Dockerfile COPY carries the whole directory, 14 files and 1,752 lines, not the 1,318 subset that silently dropped `shadcn-customize.ts`, `skills-manager.ts`, both shell scripts and the tsconfig. The second one weakened the item's own argument, which is the worse of the two.

Re-verified at head: `git grep lib/openapi` returns nothing, 293 tests, lint and types clean, `Audit, Lint, and Build` green.

nrjdalal:
Closing this in favour of shipping the review's candidates one at a time.

This PR grew to span four separate candidates (06 unwrap split, 02's vertical split, 03's `deleteSet`, 05's convention reconciliation) plus unrelated dependency churn. That is too much to review as one unit, and it couples independent changes to a single merge.

The branch `refactor/deepen-modules` is preserved; each landed piece will come back as its own PR.

PR #798, opened 2026-08-11
perf(env): load dotenv from its own module, not the one client code imports

nrjdalal:
Port of dalonic/dalonic-com#139, where this was found and merged. Same defect, same fix, measured independently here.

## The problem

`packages/env/src/lib/utils.ts` holds two unrelated things:

- the repo-root `.env` load, which needs `node:path` and `dotenv`,
- `getSafeEnv`, which is pure string work.

Every env entrypoint carries `import "@/lib/utils"` for the side effect, **including `web-next.ts`**. That module is what client components import to read `NEXT_PUBLIC_*` values, so the browserified node graph behind `dotenv` follows them into the browser: `bn.js`, `asn1.js`, AES-GCM, `sha512`, `rmd160`, Buffer and stream shims.

The browser downloads and parses all of it and never calls any of it, because the loader is guarded on `typeof window === "undefined"`.

## The fix

The side effect moves to `packages/env/src/load-dotenv.ts`, its own entrypoint with an `exports` entry and a tsdown entry. The three server env modules import it, `web-next.ts` imports nothing side-effecting, and the two build-time consumers that relied on it arriving for free now ask for it: `next.config.ts` and `generate-env.ts`.

`lib/utils.ts` held one function after the split and named nothing, so it becomes `lib/safe-env.ts`. `getSafeEnv` itself is unchanged and still re-exported from the package index.

## Measured

Landing page JS, uncompressed, summed from the chunks the page references, built from this branch and from `canary` in the same worktree:

| | Bytes | Scripts |
| --- | --- | --- |
| `canary` | 2,096,295 | 25 |
| this branch | **1,646,816** | 24 |

**449,479 bytes, 439 KiB.**

Two narrower attempts were tried in the downstream PR and neither works, so they are not worth repeating here: removing only the `web-next.ts` side-effect import saves 420 bytes, and closing only the `@packages/env` index door from inside `web/next` saves 662 bytes. Both doors lead to the same `lib/utils.ts`, so closing one leaves the other.

## Verification

The env load order changes for every server consumer, which is the kind of change that passes a build and fails at runtime.

- Full monorepo `turbo run build`, `check-types`, `lint`, `format:check` and `docs.ts --strict` pass.
- `generate-env.ts` still writes `.generated/web-env.json` (`{"isPrivate":false}`). This is the consumer most at risk, since its `web` target reaches env through a dynamic `import("@packages/env/web-next")` and so lost the transitive load.
- Standalone runtime: `bun .next/standalone/web/next/server.js` serves `/` at `200` with no validation error. `next.config.ts` does not execute there, but the client vars resolve from build-time inlining and the only runtime-read server var, `INTERNAL_API_URL`, is optional.

Downstream additionally exercised the dev stack, a DB-backed route and an auth route against the same change; those results are in dalonic/dalonic-com#139.

## Note for forks

This changes where `.env` comes from. A fork carrying its own env entrypoint, for example a `mail` target, needs `import "@/load-dotenv"` in it. The `add-package` and `codebase-map` skills say so, and the `runtime-apis` skill carries the rule the split introduced: `web-next.ts` and the package index are imported by client components, so they carry no `node:*` at all, and the `.env` load lives in `load-dotenv.ts`, which only the server targets import. `environment.mdx` (and the `llms.txt` routes generated from it) describes the same mechanism.

## Review follow-up

A review of the PR head found `environment.mdx` still naming `lib/utils.ts` as the dotenv loader, and the `runtime-apis` skill carrying only its inventory row. Both fixed on the branch, the `load-dotenv.ts` header trimmed to present behavior, and canary merged in (its #805 regeneration of `runtime-apis/index.md` conflicted with this PR's row move; resolved by regenerating the table from the merged tree, which also picked up the #805 test file that regeneration had missed).

https://claude.ai/code/session_01Sn5ADjFrSDc91U89THxy8X


nrjdalal:
**The red check is pre-existing and unrelated to this PR.**

`Audit, Lint, and Build` fails at its first step, `bun audit`, before it reaches format, lint, types or build:

```
high: nanoid: custom generators can loop indefinitely when size is zero
      GHSA-2v37-7h3g-55p8
1 vulnerabilities (1 high)
```

This branch changes no dependencies. The only `package.json` edit is the four-line `exports` entry for the new module, and `bun.lock` is untouched:

```
packages/env/package.json | 4 ++++
1 file changed, 4 insertions(+)
```

So `canary` fails the same step today. Everything after the audit gate passes locally on this branch: full `turbo run build`, `check-types`, `lint`, `format:check`, `docs.ts --strict`, plus the standalone smoke and the `.generated/web-env.json` check in the description. Both Vercel deployments are green.

The advisory needs its own PR. Downstream cleared it by refreshing `nanoid` to `3.3.17`, then later removed the override once the transitive bump made it inert, so it is worth doing deliberately rather than folding into this one.

PR #800, opened 2026-08-12
fix(deps): clear the nanoid advisory by refreshing one resolution

nrjdalal:
`bun audit --audit-level high` currently fails on **canary itself**, so it fails on every PR opened against it. [GHSA-2v37-7h3g-55p8](https://github.com/advisories/GHSA-2v37-7h3g-55p8) (high) affects `nanoid <3.3.17`, and the lockfile pinned `postcss`'s copy at `3.3.16`.

`postcss@8.5.23` declares `nanoid: "^3.3.16"`, which already admits the patched `3.3.18`, so the vulnerable copy was only a stale pin. This bumps that one resolution. No manifest change, no override, one line of `bun.lock`.

Only that copy was ever in range: `web/next` depends on nanoid directly at `^6.0.1` (used by `customAlphabet` in `src/lib/utils.ts`) and `@scalar/types` declares `^5.1.6`. Neither range admits `3.3.x`.

## Why this needed checking past the audit

Three plausible fixes give a green `bun audit` without actually fixing anything, and I hit two of them before landing this:

- **A bare `"nanoid": "^3.3.17"` override** applies to every copy in Bun, dragging the app's own runtime dependency down three majors and Scalar's down two. A nested `"postcss": { "nanoid": ... }` is not an alternative: Bun does not record nested overrides, so it resolves to nothing.
- **`bun update postcss`** moves the root copy to `8.5.26` but leaves `@tailwindcss/postcss`, `next` and `shadcn` each pinning their own `postcss@8.5.23`. One shared vulnerable copy becomes three.
- **Deleting the `postcss/nanoid` line** makes `bun audit` pass because the lockfile no longer names the resolution, while `nanoid@3.3.16` is still physically installed. A false green.

## Verification

After a clean `rm -rf node_modules && bun install`:

```
$ grep -o '"nanoid@[^"]*"' bun.lock | sort -u
"nanoid@3.3.18"   "nanoid@5.1.16"   "nanoid@6.0.1"

$ ls -d node_modules/.bun/nanoid@*
nanoid@3.3.18   nanoid@5.1.16   nanoid@6.0.1        # no 3.3.16

$ bun audit --audit-level high                       # clean
```

`format:check` and `lint` clean. `.github/notes/dependencies.md` records all of the above, including the three traps, so a returning advisory is recognised rather than re-investigated.

Carved out of #799, which is blocked on this.


nrjdalal:
Superseded by #801, which cleared GHSA-2v37-7h3g-55p8 on canary via `bun audit fix` (postcss's nanoid is at 3.3.18). The dependencies.md record of the three trap fixes is salvaged in #803 so it isn't lost.

PR #804, opened 2026-08-23
docs(audit): record which node: built-ins can move to Bun APIs

nrjdalal:
## What

A read-only audit of every `node:` import in the repo against the `runtime-apis` rule, using Bun 1.4.0's own API docs (`bun.com/llms-full.txt`) as the reference for what a Bun equivalent is. Findings are recorded in `.github/notes/audits/2026-08-23-node-to-bun-apis.md`; no runtime code changes here.

**Result**: of 33 files importing a `node:` built-in, 15 run only under Bun, and in those exactly two scripts have calls with a Bun equivalent:

| File | Calls | Bun equivalent |
| --- | --- | --- |
| `.github/scripts/ensure-remote-branches.ts` | `execFileSync` (git helper + push) | `Bun.spawnSync` with an `exitCode` check (it does not throw like `execFileSync`) |
| `.github/scripts/shadcn-customize.ts` | `execFileSync`, `readFileSync`, `writeFileSync` | `Bun.spawnSync`, `Bun.file().text()`, `Bun.write` |

Everything else stays, with the reason written down: `node:path` (Bun ships no path module), the directory half of `node:fs` (Bun's docs route `mkdir`/`readdir` to `node:fs`), `node:os`, the CLI tests (they mirror the CLI's `node:` style by rule), and `skills-manager.ts`'s `createHash`, kept deliberately because it is a line-for-line mirror of the CLI's ledger digest. Three adjacent non-`node:` candidates are assessed and declined for now (`@hono/node-server` + `ws` on Vercel, `postgres` vs `Bun.SQL`, `sharp` vs `Bun.Image` with measurements).

**Drift fixed**: `.agents/skills/runtime-apis/index.md` was a July snapshot missing six files and still classed `web/next` as a Node area. Regenerated from the current tree, `web/next` now Both, and the convertible section updated to these findings.

## Next

If the two-script migration is wanted, it is one small follow-up PR (about twenty lines), verified by `bun run test` plus a `bun run shadcn:update` end to end; the audit file is deleted once that ships or is consciously declined.

https://claude.ai/code/session_01MtdjXQt9EgfHLXb6fyfn43

### DIFF
--- a/.agents/skills/runtime-apis/index.md
+++ b/.agents/skills/runtime-apis/index.md
@@ -1,10 +1,11 @@
 # Node API index
 
 Per-file inventory of every Node built-in used in the repo, for the [`runtime-apis`](SKILL.md) skill.
-Snapshot: 2026-07-26. Regenerate with the `rg "node:..."` command in `SKILL.md`.
+Snapshot: 2026-08-23. Regenerate with the `rg "node:..."` command in `SKILL.md`.
 
-The `Runtime` column drives the rule: **Node** and **Both** files stay on `node:` (no `Bun.*`);
-**Bun** files may move a call to a `Bun.*` equivalent where one exists.
+The `Runtime` column drives the rule: **Node**, **Both** and **Build** files stay on `node:` (no `Bun.*`);
+**Bun** files may move a call to a `Bun.*` equivalent where one exists. `web/next` is **Both**: `next dev`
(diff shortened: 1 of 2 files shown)


nrjdalal:
Superseded by https://github.com/nrjdalal/zerostarter/pull/805: the findings are applied there instead of recorded, and the index refresh rides along.

PR #811, opened 2026-08-24
chore(deps): bump the catalog to latest (better-auth 1.7)

nrjdalal:
Replicates the local catalog update on a fresh worktree off canary and fixes what better-auth 1.7 breaks. Draft until the migration rollout below is approved: merging this without applying `0004_account_issuer.sql` leaves every social sign-in broken on canary, and the shared database couples the rollout to the next release.

## The bump

better-auth 1.6.23 -> 1.7.1, takumi 2.5.6 -> 2.12.0, @hono/standard-validator 0.3 -> 0.4, fumadocs 16.15.1, hono 4.13.4, oxlint 1.80, oxfmt 0.65, plus patch-level updates (posthog-js, svgo, tldts, portless, @types/*).

## What 1.7 broke, and the fixes

1. `internalAdapter.createUser` gained a required provisioning-source argument: the agent sign-in route now passes `{ method: "agent" }` (the method union is open for custom flows).
2. 1.7 keys accounts on `(issuer, accountId)` and requires `account.issuer`. Against the old schema both sides of the OAuth callback fail: `createAccount` throws `The field "issuer" does not exist in the "account" Drizzle schema` and `findAccountByKey` emits malformed SQL, so new and returning social sign-ins are all dead. The schema gains `issuer` plus the `account_issuer_accountId_uidx` unique index, and the organization plugin's `team.member_count` and `team_member.membership_key`, matching `bunx auth@1.7.1 generate` output for our config exactly.

## Migration rollout (needs approval before merge)

`0004_account_issuer.sql` adds `issuer` **nullable** and backfills it with the exact values 1.7 computes at sign-in, verified in the shipped code: `google` -> `https://accounts.google.com`, `credential` -> `local:credential`, else `local:oauth:<provider_id>`. Nullable because canary and production share one database and production stays on 1.6 until the next release; 1.6 inserts account rows without an issuer, which a NOT NULL column would reject. The follow-up (backfill stragglers, SET NOT NULL) is planned in `.github/notes/plans/account-issuer-not-null.md` and should land right after the release ships. The one soft spot in the window: an account created by 1.6 on production and signed in again on 1.7 before the follow-up backfill gets a duplicate account row, so cut the release promptly after merging.

## Verification

- `check-types` 13/13, tests 278/278, oxlint 1.80 and oxfmt 0.65 clean, full production build (web + api + cli) green.
- Disposable pglaunch database: migration chain applies both on an existing database (backfill CASE verified against seeded fake `github`/`google`/`credential` rows) and from scratch (issuer column + compound unique index present).
- Runtime on the migrated schema: `createAccount`, `findAccountByKey`, `findAccountOwnerByKey` all succeed and the unique index rejects duplicate `(issuer, accountId)` pairs.
- Real browser (agent-browser) on the worktree dev stack: homepage, sign-in modal -> Login (agents) -> dashboard authed as LocalAgent, docs page (fumadocs 16.15 + takumi 2.12 `/og` render fine). Screenshots delivered in-session; litterbox.catbox.moe has returned 403 all session, will attach here once it recovers.
- Fake seeded rows only; no real user data touched (shared Neon DB never used).

Companion PRs upstream: nrjdalal/the-next-starter#10 (registry schema + Next bump) and nrjdalal/awesome-templates#21 (sync-schemas generator), which fix the same 1.7 drift across the template pipeline.

https://claude.ai/code/session_01UQws4o1n4PHoYCKu3Yk62P


nrjdalal:
Superseded by #821: better-auth 1.7.3 dropped the issuer requirement (accounts key on providerId and accountId again), so the issuer migration here is obsolete. #821 carries the createUser provisioning-source fix and the two team columns from this branch, generated from the auth config instead of hand-written.

PR #813, opened 2026-09-04
chore(deps): adopt the cn package and zod 4.5

nrjdalal:
Adopts the new [`cn` package](https://github.com/shadcn-ui/cn) (the engine shadcn/ui migrated its whole registry to in [shadcn-ui/ui#11758](https://github.com/shadcn-ui/ui/commit/c257f688cf4de7ec10cc1be84cad29cd4631182c)) and bumps zod to [4.5](https://zod.dev/blog/zod-4-5).

## What changed

1. **`chore(deps)`**: catalog swaps `clsx` + `tailwind-merge` for `cn@^0.2.5` (same API, ~30x faster class merging), zod `^4.4.3` -> `^4.5.4`. `lib/utils.ts` re-exports `cn` so every import keeps working.
2. **`chore(ui)`**: `bun run shadcn:update` against the migrated registry. All 59 `ui/` components now `import { cn } from "cn"`; the diff is pure import swaps (+60/-97), components otherwise identical to the #806 sync.
3. **`refactor(web)`**: app components import `cn` from the package directly too, matching the registry components; `lib/utils.ts` keeps only the repo helpers (`slugify`, `isActive`). One convention instead of two. `clsx`/`tailwind-merge` remain only as transitive deps (cva, recharts, the shadcn CLI).

## Verification

- **Merge parity**: fuzzed ~44k comparisons of `cn(...)` vs the old `twMerge(clsx(...))` using 508 real class tokens extracted from this codebase (pairwise conflicts, random combos, object/array/conditional inputs): **zero output differences**.
- **zod 4.5 breaking changes**: the repo's exposed surfaces all parse identically under 4.5.4: `z.url()` env schemas (localhost and https URLs), the blog frontmatter union (`z.iso.date()`, `z.iso.datetime({ offset: true })`, `Date`), and `z.string().slugify()`. Full `next build` (132 static pages, every MDX frontmatter) green on each commit.
- **Checks**: check-types, oxlint, and the full suite (278/278) pass; each commit gated by the pre-commit full build.
- **Browser e2e** (agent-browser, 1782x972, disposable pglaunch DB with fake seeded rows): home (dark + light), docs (sidebar/TOC/collapsibles), blog post, login dialog -> agent sign-in -> dashboard, console users data table (13 rows, badges, dropdowns, selection). All render pixel-consistent with the production baseline captured at the same viewport.

Before/after screenshots are attached in [the comment below](https://github.com/nrjdalal/zerostarter/pull/813#issuecomment-5542869366).

[...]

PR #815, opened 2026-09-05
chore(deps): refresh the catalog to the latest in-range versions

nrjdalal:
`bun run clean && bun update -r && bun i` plus the full verify chain. Every catalog entry moves to the latest version its caret admits, with one deliberate exception.

## Held back: better-auth stays 1.6.23

`^1.6.23` admits 1.7.2, but 1.7 keys accounts on `(issuer, accountId)`: without the `account.issuer` migration and the `createUser` provisioning-source fix staged in #811, social sign-in breaks at runtime (proven when diagnosing the catalog bump). On the shared canary/prod database that rollout ships as its own deliberate release, so this refresh pins the lockfile at 1.6.23 (the exact-pin -> `bun i` -> deps-manager flow keeps the caret spec).

## Notable moves

base-ui 1.7 -> 1.8, react/react-dom 19.2.8, next 16.3.4, hono 4.13.7, fumadocs-core/ui 16.15.7 + fumadocs-mdx 15.4.0, takumi 2.5.6 -> 2.13.5, tanstack react-table 9.2.4 / react-query 5.102.8, posthog-js 1.427.2, shadcn CLI 4.21.0, plus the remaining catalog patch/minor bumps.

## Verification

- oxlint and oxfmt clean, check-types 13/13 (all fresh executions), 278/278 tests, full build green (132 static pages, web build cache miss so it really ran).
- Real browser on a disposable pglaunch DB: home, docs (fumadocs 16.15.7), blog, agent sign-in -> dashboard (better-auth 1.6.23 path), and /og 200 (takumi 2.13.5). Screenshots below.

| Home | Docs | Dashboard (signed in) |
| --- | --- | --- |
| Home | Docs | Dashboard |

https://claude.ai/code/session_01UQws4o1n4PHoYCKu3Yk62P

## Summary by CodeRabbit

* **Chores**
  * Updated pinned dependency versions to newer releases.


nrjdalal:
## End-to-end + Docker verification

**E2e journey** (agent-browser, 1782x972, disposable pglaunch DB, fake rows only):
sign-in dialog -> Login (agents) -> dashboard -> **created an organization** (toast, selector updated, DB row `Acme Test Org | acme-test-org-kNxE`, exercising the slugify moved in #813) -> console users **search** -> **role change** Test User 3 to member (row written, activity logged) -> **allowlist add** (`@example.com | domain` in the DB; the form correctly rejected a bare `example.com` first) -> activity log renders the "Set role" entry -> **log out** back to home.

| Org created | Role change | Allowlist | Activity |
| --- | --- | --- | --- |
| Org | Role | Allowlist | Activity |

**Docker** (`docker compose build --no-cache` then `up`, same Dockerfiles prod ships):
- The first build caught a real bug: the better-auth holdback left `bun.lock` recording the exact catalog spec while deps-manager had rewritten `package.json` to the caret, so the Dockerfile's `bun install --frozen-lockfile` refused ("the catalog in package.json changed since bun.lock was saved"). Fixed in `d043325e` (one lockfile line); reproduced and verified via a local `turbo prune` + frozen install.
- api `/api/health` 200 (`0.1.25`), web `/` 200 serving the real page.
- Full-stack DB proof: `POST /api/waitlist` through the api container wrote the row (verified via psql); the count endpoint answers by design.
- The local-only `/api/agents` route 404s in the image, matching the deployed posture.
- **Single-libc check**: OK, no glibc stack shipped in the web image.
- **Offline self-containment**: api boots and serves health under `--network=none`, so the bundle resolves everything without runtime auto-install.

CI is green on the lockfile fix (all checks pass, CodeRabbit review completed).

PR #819, opened 2026-09-06
fix: address the deepsec scan findings and record the audit

nrjdalal:
## Summary

Ran [vercel-labs/deepsec](https://github.com/vercel-labs/deepsec) 2.3.9 (agent-powered vulnerability scanner, claude-opus-4-8 at thinking level xhigh, paid through the local Claude Max login) over the trust-boundary scope: `api/hono`, `packages/*`, `.github/workflows`, the console and protected routes, and the app's auth and config libs. 59 files, about $23 API-equivalent. Every finding was re-verified by hand against the source. Nothing was rated CRITICAL or HIGH, and the auth and console surface (thirty files) produced zero findings.

This PR records the dated audit under `.github/notes/audits/` and ships the fixes that are safe to merge, with tests under `tests/` for the CLI, env and web logic changed and for the labeler's mapping; the release workflow carries none, since an `issue_comment` run executes canary's copy of the workflow file and the edit cannot be exercised from a PR. The rate limiter change was built, verified and then reverted out to keep this PR to safe changes; it is iceboxed as top priority. The `.deepsec/` scaffold stays out of the tree.

| #   | Finding                                                                           | Where                                      | Verdict              |
| --- | --------------------------------------------------------------------------------- | ------------------------------------------ | -------------------- |
| 1   | Crafted `.gitpickignore` deletes outside the root on in-place convert             | `packages/cli/src/convert.ts`              | Shipped              |
| 2   | Project name interpolated unescaped into the generated `site.ts`                  | `packages/cli/src/templates.ts`            | Shipped              |
| 3   | Rate-limit key trusts client forwarding headers and falls back to a random key    | `api/hono/src/middlewares/rate-limiter.ts` | Icebox, top priority |
| 4   | `/release` comment builds and publishes a PR-selected commit with no origin check | `.github/workflows/cli-release.yml`        | Shipped              |
| 5   | Top-level `id-token: write` and write scopes reach the test job                   | `.github/workflows/cli-release.yml`        | Shipped              |
| 6   | Fork PR file paths mint arbitrary `@<dir>` labels                                 | `.github/workflows/auto-labeler.yml`       | Shipped              |
| 7   | Brand name used as a `replaceAll` replacement string                              | `packages/cli/src/skills.ts`               | Shipped              |
| 8   | `action in ACTION_LABELS` walks the prototype chain                               | `web/next/src/lib/activity.ts`             | Shipped              |
| 9   | Docker host rewrite replaces the first `localhost` anywhere in the URL            | `packages/env/src/db.ts`                   | Shipped              |
| 10  | Version strings shell-interpolated into inline scripts (inputs trusted)           | `.github/workflows/auto-release.yml`       | Plan                 |
| 11  | Actions on mutable major tags in token-bearing workflows                          | two workflows                              | Icebox               |
| 12  | Public `/og` renders a PNG per distinct query string                              | `web/next/src/app/og/route.tsx`            | Accept               |
| 13  | The `*.vercel.app` `/api` rewrite bills a Vercel hop, not the client (pre-existing, found while verifying) | web rewrite path | Icebox |

## Why the rate limiter is not in here

The fix as built keyed on the socket peer and filtered headers by hand. Researching `@arcjet/ip` showed that is the wrong model: the library resolves a connection address first and then only the headers its `platform` option names, with `proxies` (including its shipped `cloudflare` service) for anything in front, which is how arcjet's own Bun and Node adapters call it. Adopting that is a design change with a live finding behind it, so it is parked in `plans/rate-limit-ip-resolution.md` as the top icebox item, with everything the reverted work proved recorded there, and this PR stays to changes that carry no such decision.

## Also in here

- Three older test tsconfigs gained the `exclude` the others carry, at the owner's request during review, so every test tsconfig shares one shape.
- Item 10 is recorded in the workflow-tooling-consistency plan; items 3, 11 and 13 are iceboxed (`plans/rate-limit-ip-resolution.md`, `plans/action-sha-pinning.md`, `plans/rate-limit-rewrite-bucket.md`, the plans index, and #707) and the audit points at them.

## Verification

Every claim above was exercised against a real system; transcripts and screenshots are in the comments below (the limiter transcripts there describe the reverted work).

- `bun run test`: 297 pass, 0 fail (19 new tests across six test files, one of which lifts the labeler's mapping out of the workflow and runs it against the checkout). `bun run check-types`: 13/13. `bun run lint` and `bun run format` clean. actionlint reports no errors on either edited workflow; its eight shellcheck notes on the release script are pre-existing lines, identical in count and rule to canary.
- Release path: from a fresh clone, `bun install --frozen-lockfile --ignore-scripts` followed by the CLI's `bun run build` produces a working binary, so the publish job needs no lifecycle script.
- Docker images built `--no-cache` from this branch: offline self-containment (no auto-install), single-libc check OK, and a waitlist signup from the host landed in a disposable Postgres that the container could only reach through the rewritten `host.docker.internal` URL.
- CLI: the built binary run non-interactively in place on fake checkouts refuses a `.gitpickignore` that climbs out (nothing removed, whichever entry is bad: every entry is checked before any is removed), writes `name: "Ac\"me"` for a directory named `ac"me` (the module imports), and rebrands a skill literally for `Acme $& Co`.
- Labeler: the tracked top-level directories come from `git ls-tree` into a set, so a pathspec glob such as `*` cannot match everything; the automated test covers a novel directory, `*`, `[a-z]*` and an installed `node_modules` (all `@misc`) and the tracked ones (their labels).
- Browser: seeded a disposable database with actions `constructor`, `toString`, `fork.custom`; before, those rows render a blank Action cell and Next reports an issue; after, the codes render as text. Responsive and theme N/A (a label lookup, no layout or style change).


nrjdalal:
**Rate limiter, verified against the running Bun server** (`PORT=4123 bun src/index.ts`, requests straight at the port, so the peer is loopback):

```
=== direct, no forwarding header (loopback peer, internal): expect no RateLimit headers
HTTP/1.1 200 OK
HTTP/1.1 200 OK
=== direct, x-forwarded-for 1.1.1.1 twice: expect limit 60, remaining 59 then 58
HTTP/1.1 200 OK RateLimit-Limit: 60 RateLimit-Remaining: 59
HTTP/1.1 200 OK RateLimit-Limit: 60 RateLimit-Remaining: 58
=== direct, x-forwarded-for 9.9.9.9: separate bucket, remaining 59
HTTP/1.1 200 OK RateLimit-Limit: 60 RateLimit-Remaining: 59
=== 1.1.1.1 until throttled
429 after 61 requests: {"error":{"code":"TOO_MANY_REQUESTS","message":"Too Many Requests"}}
=== internal peer never throttled: 70 bare requests
  70 200
```

So `getConnInfo` reads the peer from the real server object, a private peer defers to the forwarded client, a private peer with nothing forwarded is skipped, and the budget accumulates per client instead of resetting on a random key. The public-peer path (a forged header ignored) is covered by the unit test with a stubbed server, since a local socket cannot present a public address.

nrjdalal:
**Full verification pass** (every claim in the description re-run against a real system; the earlier limiter comment covered the direct Bun port).

**Rate limiter**

- Through portless (loopback proxy, loopback client): no `RateLimit-*` headers, so internal and skipped. A client-supplied `x-forwarded-for` through portless is honored (`59`, `58`): the peer is loopback, which is the private-network case the docs scope to a proxy that sets the header.
- `docker compose` (both images built `--no-cache` from this branch, real env, disposable Postgres on the host):
  ```
  host -> api :4000, no header (peer 192.168.97.1, private):  200, no RateLimit headers   x2
  host -> api :4000, x-forwarded-for 1.1.1.1:                 RateLimit-Remaining 59, 58
  web container -> api:4000 (sibling, private peer):          200, no RateLimit headers
  70x GET /dashboard through the web (server-side session reads to the api): 70x 307, zero 429
  ```
- Vercel preview of `3c415c00` (`/api/health` reports `0.1.26-3c415c0`): no header twice, then forged `1.1.1.1`, then forged `9.9.9.9`: remaining `58, 57, 56, 55`, one bucket.
- Production, pre-PR code, same probe for comparison: direct and via the `zerostarter.dev/api` rewrite share one bucket (`58 … 52`). Via the preview web's `/api` rewrite (which reaches the production API) the bucket differs, so that path bills a Vercel hop, not the client: pre-existing, recorded as item 13 in the audit.
- Library facts behind the design: `@arcjet/ip` classifies `8.8.8.8`, `1.1.1.1`, `2606:4700::1111` as global and `10/8`, `172.16/12`, `192.168/16`, `100.64/10`, `169.254/16`, `127.0.0.1`, `::1`, `fe80::`, `fc00::` and `::ffff:8.8.8.8` as nothing. Bun on `0.0.0.0` (the API's default) reports IPv4 peers as `127.0.0.1`; on `::` it reports them as `::ffff:127.0.0.1`, hence the unwrapping commit.

**Env: Docker host rewrite** (fix 9). Inside the api container the raw env is `POSTGRES_URL=postgres://***@localhost:59546/postgres` with `INTERNAL_API_URL=http://api:4000`. `POST /api/waitlist` from the host answered `{"data":{"message":"ok"}}` and the row landed in the host's disposable Postgres, which the container can only reach as `host.docker.internal`.

**Docker image checks** (docker-test skill): offline self-containment of the api image with `--network=none`: health OK, `0` auto-install cache entries. Single-libc check on the web image: `OK`.

**CLI** (the built `packages/cli/dist/bin/index.mjs`, non-interactive in-place `init` on fake checkouts):

```
A. .gitpickignore = "../victim-dir/victim.txt\nLICENSE.md"
   exit 1: .gitpickignore entry "../victim-dir/victim.txt" is not inside the project; the in-place converter only removes paths under the root.
   victim intact: keep | LICENSE.md still present: yes      (nothing removed before the refusal)
B. directory ac"me          -> site.ts:  name: "Ac\"me",   description: "Ac\"me is just getting started. ..."
   bun import of site.ts   -> site.name = "Ac\"me"
C. directory "Acme $& Co"   -> SKILL.md: description: Start the Acme $& Co dev stack.  /  Acme $& Co runs here.
```

**Labeler**: `getLabelForFile` extracted from the workflow and run on this checkout: `zzz-spam/x.txt -> @misc`, `evil/../../x -> @misc`, `tests/... -> @tests`, `.agents/... -> @.agents`, `.github/workflows/x.yml -> @workflows`, `.github/scripts/x.ts -> @scripts`, `api/hono/src/index.ts -> @api/hono`, `docs/x.md -> @misc`, `README.md -> @misc`, `bun.lock -> @dependencies`, `Dockerfile -> @docker`.

**Release workflow**: `bunx json head.repo.full_name` and `head.sha` return the fields from a PR payload. actionlint (1.7.7) reports no errors; its eight shellcheck notes are the same pre-existing lines as on canary. The `/release` comment path runs from the default branch's workflow file, so it cannot be exercised from this PR.

**Web** (agent login, disposable database seeded with actions `constructor`, `toString`, `fork.custom`, `role.change`), same viewport 1782x972:

Before (canary's `in` lookup): the `constructor` and `toString` rows render with a blank Action cell and Next's dev overlay reports 1 issue. After: the codes render as text. Responsive and theme: N/A, a label lookup with no layout or style change.

**Suite**: `bun run test` 303 pass, 0 fail (25 new); `bun run check-types` 13/13; lint and format clean; CI green on `3c415c00`.

activity-before

activity-after

nrjdalal:
**Review follow-up** (standards 4 hard + 6 judgement, spec 7). History was rewritten for the commit types, so the branch is force-pushed; nineteen commits now.

Standards
- `ci` type: the two commits are reworded to `fix(ci)` (gh-commit skill; `changelog.config.json` drops `ci`).
- Undecided items: 11 and 13 are iceboxed per the skill (`plans/action-sha-pinning.md`, `plans/rate-limit-rewrite-bucket.md`, Icebox links in the plans index, two boxes on #707); the audit points at them instead of holding them.
- runtime-apis inventory: rows for `convert.ts`, `convert.test.ts` and the labeler updated.
- Alphabetical members: `RateLimiterConfig`, its destructuring and the `rateLimiter({...})` call are A to Z.
- `onVercel` is exported from `lib/server` and imported; the decision is memoized per request (`WeakMap` on the raw `Request`), so `findIp` runs once; `docker.mdx` names `lib/postgres-url.ts`; the tooling plan no longer counts its threads; the two outside tests share a victim-directory helper; the `skills.ts` note joined its header comment and the limiter comments state the Vercel point once.
- Left as is: the two new `bunx json` calls follow the release script's existing unpinned calls, which the tooling plan already tracks pinning together.

Spec
- Header set (the main finding): only `x-forwarded-for` reaches `@arcjet/ip` now; a private peer plus `x-real-ip` / `x-client-ip` / `true-client-ip` / `forwarded` alone is internal, and no peer plus any of them is `ip:unknown`, both tested. Description and docs say so.
- Absolute entries: refused outright, inside or outside the root, with a test for the inside case.
- Labeler: a directory earns a label only if `git ls-files` shows tracked files under it, so the install's `node_modules` cannot mint one; verified locally against tracked, installed and untracked directories.
- Code comment and mdx no longer claim the rewrite is covered unconditionally; item 13 carries the observed caveat.
- Shared `ip:unknown` bucket is now in the mdx.
- tsconfig normalization: kept, requested by the owner mid-review ("every single [config] should match the other configs"); the body says so.

Checks: 304 tests, 13/13 type tasks, lint, format and actionlint (no new findings) all green locally.

nrjdalal:
**Second-pass follow-up** (standards 3 hard + 5 judgement, spec 5). Two commits were split for atomicity and seven added, so the branch is force-pushed again; 28 commits, 30 files.

Standards
- Doc callout: now says the library is handed the socket peer and `x-forwarded-for`, nothing else.
- Fallback list: `forwarded, true-client-ip, x-client-ip, x-real-ip` in the comment, the test and the doc, in its own `style:` commit.
- Non-atomic commits: `e9360cdd` is now `fix(cli)` plus `style(cli)`; `01801a2c` is now `docs(audits)` (the description corrections and the plan wording) plus `docs(plans)` (the icing). Trees identical before and after.
- Decider memo: kept as a per-limiter `WeakMap`, with the reason in the comment and a test: the global and the per-user limiter both see one request and must not read each other's answer, which a single context variable would hand them.
- `convert.ts` refusals share one message builder; the runtime-apis snapshot date is stamped.
- Left as is: the two `bunx json` calls follow the release script's existing unpinned ones, tracked together by the tooling plan; `onVercel` stays a raw `process.env.VERCEL` read in `lib/server.ts`, pre-existing and now single-sourced, since `VERCEL` is not part of the env schema.

Spec
- Once-per-request: `rateLimitDecider` is exported and tested (resolver runs once for repeated asks, a fresh request resolves again, two limiters keep separate answers).
- Labeler test: `tests/github/workflows/auto-labeler.test.ts` lifts the mapping out of the YAML and runs it against the checkout.
- Glob pathspec: the tracked top-level directories are read once with `git ls-tree --name-only -d HEAD` into a set; `*`, `[a-z]*` and `**` map to `@misc`, tested.
- "Nothing removed": every entry is validated before any removal; a good entry ahead of a bad one, and a glob after a good one, both leave the checkout untouched, tested.
- Vercel server-side reads: the body now says they are skipped off Vercel only, and on Vercel key on what Vercel stamps (item 13).
- tsconfig scope: the owner's request was made mid-review in the session ("whenever you are writing a TS config, there is a standard pattern that every single repo that follows it should match the other configs"), quoted here so it is on the record.

Checks: 311 tests across 31 files, 13/13 type tasks, lint, format and actionlint (unchanged, eight pre-existing shellcheck notes) all green locally.

nrjdalal:
**Third-pass follow-up** (standards 4 hard + 8 judgement, spec 3). Five commits added; 33 commits.

Standards
- Inventory: `tests/github/workflows/auto-labeler.test.ts` has its row, and the test now imports only `node:module` and `node:path`; a sweep of every file with a `node:` import against the inventory finds no other gap.
- Audit line 31 no longer counts commits per fix.
- The prototype-name list in the activity test is A to Z.
- The labeler test reads the workflow with `Bun.file`, like its sibling, drops the probe directory it wrote into the repo root (`node_modules` already stands in for untracked disk content), and restores the working directory after the run.
- Left as is, with reasons: the three judgement-call commit bundles (count edit with audit corrections, two doc syncs, the `refuse` helper with its fix) are each one editing pass over one concern and a fourth history rewrite on a reviewed PR costs more than it buys; the workflow test tsconfig keeps its truthful first line since a workflow has no config of its own; the limiter's three exported helpers exist so the pure logic is unit-tested, the same reason `access.ts` exports its predicates; the resolver pair stays two parameters, mirroring `RateLimiterConfig`; the `new Function` evaluation is deliberate and commented; `bunx json` and `process.env.VERCEL` remain as before.

Spec
- PR body no longer states a tsconfig count.
- `rate-limiting.mdx` item 4 and audit item 3 qualify the skipped internal traffic as off Vercel, where Bun owns the socket.

Checks: 311 tests across 31 files, 13/13 type tasks, lint and format clean locally.

nrjdalal:
**Fourth-pass follow-up** (standards 1 hard + 9 judgement, spec 4). Two docs commits; 35 commits.

Standards
- The item-13 plan lists its candidate headers A to Z.
- The plans index gloss for the tooling plan now names the version pass-through it carries.
- Left as is: the `node:fs` removal and the probe-directory removal share one commit because they share the import line they delete; the cwd swap is restored and unavoidable, since the workflow's `execFileSync` takes no cwd.

Spec
- PR body and audit no longer claim one commit per fix, and both say which files carry tests: every CLI, API, env and web module touched plus the labeler's mapping, with the release workflow carrying none because its comment path only runs from the default branch.
- The body's labeler case list matches the test (`node_modules` stands in for untracked disk content).
- "The other four" is now "the others".
- On the commit body's "as its sibling does": `tests/github/scripts/console-roles.test.ts` reads its subject with `Bun.file().text()`, which is the sibling meant; `ensure-remote-branches.test.ts` is the one using `Bun.spawnSync`.

No code changed in this pass.

nrjdalal:
**Fifth-pass follow-up** (standards 0 hard + 14 judgement, spec 3). Doc-only; a461d129 split into two commits and one audit commit added, so the branch is force-pushed; 37 commits, tree identical apart from the audit sentence.

Standards
- a461d129 is now two commits, one per fix.
- The audit's module list reads API, CLI, env, web.
- Left as is: the fallback-header set is mirrored in the comment, the test and the doc by nature; the earlier bundles and the other carried-over judgement calls stand as previously reasoned.

Spec
- Body and audit now say the tests cover the API, CLI, env and web logic changed, naming `server.ts` and `db.ts` as the two touched modules without a mirror (one exports a flag, the other calls the tested helper).
- The release-workflow sentence now gives the real reason: an `issue_comment` run executes canary's copy of the workflow file, so the edit cannot be exercised from a PR; the path itself checks out and publishes the PR head, as the commit body already said.
- "And every subject config" is dropped: `.github/scripts/tsconfig.json` has no `exclude`.

No code changed in this pass.

nrjdalal:
**Sixth-pass follow-up** (standards 0 hard + 10 carried, spec 1). Message-only: the tsconfig commit's body no longer says the subject configs all carry the `exclude` (`.github/scripts/tsconfig.json` does not), and no longer claims the two GitHub test configs inherit it; tsc's default exclude is what makes the change a no-op for it. Tree unchanged, branch force-pushed; 37 commits.

The carried judgement calls stand as previously reasoned. No code changed in this pass.

nrjdalal:
**Scope change: the rate limiter is reverted out and iceboxed as top priority.** Three commits, no history rewrite; 40 commits (the audit note followed in its own commit).

- `api/hono/src/middlewares/rate-limiter.ts`, `api/hono/src/lib/server.ts` and `docs/manage/rate-limiting.mdx` are byte-identical to canary again, and the limiter test is removed. The finding (item 3) therefore stands on canary; the audit says so.
- Why: researching `@arcjet/ip` showed the built fix used the wrong model. The library resolves a connection address first and then only the headers its `platform` option names, with `proxies` (including the shipped `cloudflare` service) for anything in front; arcjet's own Bun and Node adapters call it exactly that way, and `@arcjet/env` detects the platform. Hand-filtering headers would also strip `cf-connecting-ip`, and dropping `platform: "vercel"` rested on a probe of the old code. That is a design decision, not a safe fix, so it moves to `plans/rate-limit-ip-resolution.md` (first in the Icebox index and on #707) with everything the reverted work proved.
- What stays is the set with no such decision behind it: the three CLI fixes, the env host rewrite, the activity lookup, the labeler set, the release-origin gate, the permission scoping and the `--ignore-scripts` install. For the last one, a fresh clone installed with `--frozen-lockfile --ignore-scripts` builds and runs the CLI binary, so the publish job needs no lifecycle script.

Checks: 297 tests across 30 files (the limiter's fourteen gone), 13/13 type tasks, lint and format clean locally.

PR #822, opened 2026-09-06
feat(web): render vercel speed insights only when deployed on vercel

nrjdalal:
Replaces #812, which targeted `main` instead of `canary`, was `CONFLICTING` so its checks never ran, and mounted `<SpeedInsights />` unconditionally.

## Why conditional

Speed Insights posts to `/_vercel/speed-insights/vitals`, a route only Vercel serves. Mounting it unconditionally means every Docker and self-hosted deploy of the starter loads a beacon that fires requests which can never arrive. The root layout gates it on `process.env.VERCEL`, the same variable `next.config.ts` and `api/hono/src/lib/server.ts` already branch on.

No environment variable and no key: `VERCEL` is injected by the platform rather than configured, so it stays out of `@packages/env` and `.env.example`.

## What the gate does, precisely

Off Vercel the component is never rendered, so the beacon script never loads and no request is made. The package's own module is a static import, so it still ships in the client bundle either way. Nothing executes it. A dynamic import would remove it from the bundle too, which the docs note, but a wrapper component is out of proportion to what it saves.

## Why the layout and not a provider

`SpeedInsights` is a `"use client"` component, and Next inlines only `NEXT_PUBLIC_*` variables into client bundles. The same check inside `providers.tsx` would read `undefined` in the browser and silently never render, even on Vercel. Tested rather than assumed, in a comment below: SSR HTML carried 0 references, the live DOM none, and `process` is undefined in the browser.

It is also not a provider by shape: it takes no children and supplies no context, it is a leaf that injects a script.

## Verified

Both branches driven in a real browser against `web/next`, querying the live DOM for the injected beacon.

| Run | Injected beacon script |
| --- | --- |
| `VERCEL=1 bun run dev:app` | `https://va.vercel-scripts.com/v1/speed-insights/script.debug.js` |
| `bun run dev:app` | none |

The debug variant is what the package loads when `NODE_ENV=development`; production on Vercel loads the real script, which the preview deployment on this PR serves. The beacon itself has no visual surface, so the DOM query is its evidence; screenshots of the docs section, which is the visual part of this change, are attached in a comment.

`bun run format:check`, `bun run lint` and `bun run check-types` all pass, the last on real cache misses across all 13 tasks.

## Docs

- `manage/analytics.mdx` gains a Speed Insights section: why it is conditional, why the check cannot move into a client component, what the gate does and does not do, how to narrow it without also silencing `canary` (which deploys as a preview), and how to remove it.
- `deployment/vercel.mdx` gains a short section and a cross-link.
- Speed Insights named in the four stack lists that previously read as PostHog alone: the README Analytics bullet, the architecture feature line, the roadmap shipped list, and the llms preamble in `site.ts`.
- `docs.config.ts` description updated, which is what syncs the MDX frontmatter.

## Dependency

`@vercel/speed-insights` added to the root catalog at `^2.0.0`, the current latest, as a caret range per AGENTS.md. `deps-manager.ts` left it untouched on postinstall, confirming the spec is already in the required form.


nrjdalal:
Tested the provider placement rather than assuming, since it is the obvious first instinct.

Moved the check into `providers.tsx` (`{process.env.VERCEL === "1" && <SpeedInsights />}`), removed it from the layout, and ran the app with `VERCEL=1`:

| Check | Result |
| --- | --- |
| SSR HTML references to the beacon | 0 |
| Injected script in the live DOM | none |
| `process` in the browser | `undefined` |

It silently never renders, even on Vercel. `providers.tsx` is `"use client"`, and Next inlines only `NEXT_PUBLIC_*` into client bundles, so the check evaluates against nothing. There is no error, which is what makes it worse than a failure.

Making it work from a provider would need either `NEXT_PUBLIC_VERCEL_ENV` (a Vercel dashboard setting a fork would have to know to enable, so no longer zero-config) or the boolean threaded down as a prop from a server component. The prop version works but is strictly worse: the client reference then ships in every self-hosted bundle, which is the thing the gate exists to prevent.

Reverted; the branch is unchanged from the reviewed commit.

nrjdalal:
Review addressed in b8f39279. Screenshots of the docs section attached below, which is the visual part of this change.

si-docs

si-callout

nrjdalal:
Point-by-point on the review.

**Fixed**

- **Overstated bundle claim.** You are right and I was wrong. Off Vercel the component is never rendered and the beacon never loads, but the import is static, so the package module ships in the client bundle regardless. The docs now say exactly that, and point at a dynamic import for anyone who wants it out of the bundle too. I chose rewording over the dynamic import: the module is about a kilobyte, and a wrapper component to shave that is machinery out of proportion to the gain. The PR body is corrected too.
- **Narrowing advice would have silenced canary.** Good catch, that would have quietly cost you canary's data. The callout now says `VERCEL_ENV === "production"` drops canary along with PR previews, and shows the branch test `migrate-on-deploy.ts` already uses for exactly this distinction.
- **Doc-sync sweep.** Speed Insights added to all four: the README Analytics bullet, the architecture feature line, the roadmap shipped list, and the llms preamble in `site.ts`.
- **Comment length.** Cut from sixty words to one sentence. It also no longer implies `next.config.ts` uses the same literal test, which was the half-wrong part of the duplication note.
- **Removal steps** now include the comment line.
- **Screenshots** attached in the comment above.

**Not changed, with reasoning**

- **Duplication with the api check.** Left alone. Two workspaces, one line each, and a shared helper would have to live in `@packages/env`, which is where this deliberately does not belong.
- **`VERCEL` undocumented in the environment and Docker pages.** Agreed there is no new obligation, and adding it would suggest it is something you set.
- **Module-level `onVercel` baking at build time.** Correct reading, and I agree nothing diverges today. Worth leaving as a known shape rather than defending against a build that cannot currently happen.

Your review worktree at `.claude/worktrees/review-822` is untouched; remove it when you are done with it.

nrjdalal:
Second round addressed in bebfedca. CI green, `CLEAN`.

**The copy nit was the real one, and you called it right.** I checked Vercel's current quickstart rather than take either page's word for it. It has four steps: add the package, add the component, deploy, view the data. No enable step anywhere, and it explicitly notes that deployments already include the collection routes under `/_vercel/speed-insights/*`. So `architecture.mdx` was correct with "automatic" and the other two pages had invented a step that would have left someone hunting the dashboard for a toggle that does not exist. Both now say there is no enable step, and that the dashboard is where the numbers show up rather than where you switch it on.

**Volatile count removed.** "About a kilobyte" is gone from the docs and the PR body. Your measurement is the reason it should not be there: it only holds at gzip granularity and moves with each release. The sentence still makes its actual point, which is that nothing executes the module.

**Intro reframed.** You were right that "all three are optional ... you opt in" was false for this one. It now says PostHog and the feedback link are opt-in, and that Speed Insights works the other way round: no variable turns it on, and removal is how it goes away.

**Codebase-map gap closed.** Taken even though it predates this PR, since this PR is what makes it material by adding the third direct read. The env row now carries the carve-out for platform-injected variables (`VERCEL`, `VERCEL_ENV`, `VERCEL_GIT_COMMIT_*`): read at the point of use, in neither the schema nor `.env.example`. The skill's frontmatter is untouched, so the AGENTS.md table is unchanged.

Nothing outstanding from either round that I know of.

nrjdalal:
Round four addressed in 521c0ef3. CI green, `CLEAN`.

**Description trimmed**, the one item worth acting on. Measured across all 33 descriptions in `docs.config.ts` before changing anything:

| | Before | After |
| --- | --- | --- |
| This description | 156 | 91 |
| Longest in the config | 156 (this one) | 137 (data tables) |
| Median | 83 | 83 |

156 also sat right where a search snippet truncates, so it was losing its tail in two places at once. It now reads "Optional PostHog analytics and a feedback link, plus Speed Insights on every Vercel deploy." That keeps what round three was about: "optional" attaches to PostHog and the feedback link, and Speed Insights is named as arriving with every deploy rather than as something you switch on. The intro carries the full version. Frontmatter regenerated through `docs.ts` rather than hand-edited, so it stays byte-identical and unquoted.

**Both other calls taken as no-change, agreeing with your reasoning.** The carve-out appearing in both the codebase-map row and the environment page is the doc-sync skill working as intended, since skills and docs are separate surfaces that each have to stand alone. And "with one exception" does read as an example rather than an exhaustive claim, so tightening it would cost more words than it buys, particularly given `GIT_SHA` and the portless URL are different in kind from the platform variables the sentence is about.

Nothing outstanding across all four rounds that I know of. Both worktrees, yours at `.claude/worktrees/review-822` and mine at `.claude/worktrees/speed-insights`, can go once this merges.

PR #826, opened 2026-09-06
feat(ci): the release number boards with the code

nrjdalal:
## Summary

Production built the v0.1.27 release merge as `0.1.26-995fd28`: the bump lands on `canary` only after the merge, so `main` always carried the previous number. This PR decides the number *before* the merge, on every human push to `canary` while a release window is open, and puts it in the tree, so previews show the upcoming number and production shows the released one on the merge itself. Nothing flows from `main` back to `canary`; the direction is unchanged.

Three rules, in one script (`packages/scripts/src/release-version.ts`, `bun run release:version`):

1. **What a window has earned** is changelogen's bump applied to the last tag's version over the window's commits. A missing tag counts as `v0.0.0`, so a fresh fork's first number is honored. A window earns nothing when it holds only `ci(changelog)` and `ci(version)` commits, or when changelogen's own changelog for it has no entry (it drops `ci` per `changelog.config.json` and non-breaking `chore(deps)` by itself), which is exactly the pair of gates auto-release applies before it releases.
2. **Compute from the tag, not from the file.** changelogen bumps whatever `package.json` holds (measured: a pre-set 0.1.28 becomes 0.1.29), so the script runs it against a copy set to the tag's version and puts every file back. changelogen's CLI reports its own errors and still exits 0, so the script reads success from what the run left behind (the version moved past the tag's and a `## v` section for it heading the changelog) and fails loudly otherwise: a broken `changelog.config.json` is an error, not an empty window.
3. **Forward only, never past a human.** The tree moves to `max(tree, earned)`: a hand-set 2.0.0 stays, a hand-set 0.1.28 in a breaking window is lifted to the 0.2.0 it earns, and a tree below the last tag is refused loudly.

`auto-canary-into-main.yml` runs the script with `--write` after opening or finding the draft release PR, on the `canary` ref only, and commits `ci(version): bump to vX.Y.Z`; a push that loses a race with a human push is told apart from a real failure by comparing the remote against the commit the run started from. `auto-release.yml` and the CLI's canary prerelease preview take their target from the same script, captured before it is piped so a script failure stops the step instead of becoming an empty version (the preview installs first, since the script runs the scripts package's own locked changelogen), and auto-release counts `ci(version)` as mechanical in its skip gate. Taking the target from the script also replaces the old rewrite that only ran once a tag existed, which fixes a first-release surprise: a fork that hand-set 1.0.0 before its first tag used to ship 1.0.1.

**One more thing the browser run found:** the root `package.json` was not a turbo cache input, so a version-only commit replayed a cached `@packages/env` build and the baked number stayed old. `turbo.json` now declares it as an input of that one build (`$TURBO_ROOT$/package.json`), the only task that reads it, rather than a global dependency that would miss every cache on any root edit; the apps rebuild behind it through their dependency hashes. Measured locally: a version change is a cache miss for the env build without `--force`, the API build's hash changes with it, and restoring the file is a hit again.

## Measured

changelogen 0.6.2 with this repo's config, in a throwaway checkout: at 0.x a `feat` is a patch (0.1.27 to 0.1.28; 0.0.99 to 0.0.100, which is why the 0.1.0 release was a hand-set) and a `feat!` a minor (0.1.27 to 0.2.0); past 1.0 the usual semver (1.4.2 to 1.5.0 / 1.4.3 / 2.0.0); a pre-set 0.1.28 becomes 0.1.29 and a hand-set 2.0.0 becomes 2.0.1 unless rewritten, which rule 2 and auto-release's existing rewrite handle.

## Every stage

- **`init`**: version 0.0.0, canary and main at one commit, no tag; both workflows and `changelog.config.json` ship with the fork. Unchanged.
- **First push**: no PR while main is missing or canary is level with it, so no bump. Unchanged.
- **First window**: the tree moves to 0.0.1 (or 0.1.0 with a breaking commit) from the v0.0.0 baseline; previews show it.
- **First release**: ships exactly that; production shows it on its first build.
- **Steady state**: one merge per release, number already aboard.
- **Minor or major**: the same hand-set PR as today; it is ahead of the window so it stays.
- **A fork syncing later**: picks up the workflows; the next window is the first moved.

## Edge cases

- The bump is pushed by the workflow token, which starts no further workflows: no loop, and the bump commit carries no checks of its own; the next human merge runs them all. Main's ruleset requires a review, not status checks; a fork that requires checks on main sees its release PR wait for one more merge (now in the release skill's notes).
- A window of only dependency refreshes titled `chore(deps)` earns nothing and moves nothing; titled `build(deps)` it earns a patch and releases, as today.
- A tag already present at the computed number stops auto-release loudly, as today.
- A clone whose `v*` tags exist but are not reachable from HEAD (a shallow clone without them) stops the script loudly instead of computing from v0.0.0; only a repository with no tag at all is the fresh-fork start.
- `bun.lock` lists workspace names without versions, so a version-only commit installs frozen; every changelog commit already proves it.
- The build-graph image and the changelog still ride the post-merge commit; the CLI still publishes from the tag with its own package at 0.0.0.

## Verification

- 20 tests in `tests/packages/scripts/src/release-version.test.ts`: the pure decision, and the script against real throwaway repositories carrying this repo's changelog config, shaped like each stage (no tag, feature window, breaking mid-window, mechanical-only and dropped-types windows, already moved, hand-set major with files put back, below the tag), plus a broken changelog config read as a failure, the entry point run as a subprocess with and without `--write`, the version field found at the top level whatever the file's spacing or nesting, and tags that HEAD cannot reach stopping the script.
- Dry run on canary's real window (only the changelog commit since v0.1.27): `moved: false`. On this branch, whose own `feat` commits open a window: `next: 0.1.28`, tree untouched afterwards.
- All three workflows parse as YAML; `bun run check-types` 13 of 13; lint and format clean; `bun run test` 318 pass, 67 skip, 0 fail.
- The squash-merge of this PR is itself a push to `canary` that runs the new workflow, and its own `feat` commits open a window, so the first live run moves canary to 0.1.28 right after merging.

Browser screenshots of the surfaces that display the number, before and after the tree moves, are in the comment below.

## Summary by CodeRabbit

- **New Features**
  - Added `release:version` to preview or apply the next release version.
  - Release automation now consistently determines versions from earned changes, existing versions, and release history.
  - Canary updates can automatically advance the package version before release.

- **Documentation**
  - Added usage guidance and JSON output details for the release-version command.
  - Updated release documentation to explain versioning behavior for pre-1.0 and post-1.0 releases.

- **Chores**
  - Improved release consistency across standard and CLI release workflows.


nrjdalal:
## The surfaces that show the number, before and after the tree moves

The branch's own dev stack on a disposable database, viewport 1782x972. "Before" is the tree as it is on canary today (`package.json` at 0.1.27 while this branch's commits have already earned 0.1.28). "After" is the same stack once `bun run release:version --write` moved the tree, which is what the draft-PR workflow does on every human push to canary. Same SHA in both, because only the number changed. No UI changed in this PR; responsive and theme checks are N/A.

1. `/api/health` before: the old number with the current SHA, the shape production showed after v0.1.27.

Health before

2. The dashboard sidebar before, reading `v0.1.27-52629fb2`.

Sidebar before

3. `/api/health` after the move: `0.1.28-52629fb2`, the number the release will carry.

Health after

4. The dashboard sidebar after, reading `v0.1.28-52629fb2`.

Sidebar after

Getting the "after" to show at all is what surfaced the turbo finding in the PR: without `--force`, the rebuilt env package replayed a cached build and still said 0.1.27, because the root `package.json` was not a cache input. It is now.

PR #828, opened 2026-09-06
fix(scripts): compute the release number with changelogen's library

nrjdalal:
## Summary

`release-version.ts` no longer runs changelogen's command. It uses the package as a library, which `packages/scripts` already declares: `loadChangelogConfig` reads `changelog.config.json`, `getGitDiff` and `parseCommits` read the window's commits, and `determineSemverChange` gives the semver reading of the ones the command would keep. Nothing is rendered, so no author is looked up over the network, and nothing is written, so `package.json` and `CHANGELOG.md` are never touched during a decision.

Why: running the command against a copy of the tree was the root of every guard the script had grown since #826. The `.bin` shim Windows does not have, the command's exit-0 error handling, the version-and-heading success check, the temp writes and restores, and finally the author lookup on ungh.cc that stalled the v0.1.28 publish leg on macOS (Bun killed the child at the 30s test limit; a re-run passed). An earlier revision of this PR blackholed that lookup through the proxy variables; this one removes the reason for it.

**Four rules are mirrored, one line each**, because the command applies them outside the functions it exports: the type is lowercased, since the parser is case-insensitive; commits of a disabled type or a non-breaking `chore(deps)` are dropped before anything renders (the scope keeps its case in both, so `Chore(Deps)` is kept by both and they agree); a window with kept commits but no semver-bearing type bumps a patch; below 1.0 a major counts as a minor and a minor as a patch. A window that keeps no commit earns nothing, which is exactly the entry auto-release's content gate demands. One deliberate difference: the type must be the config's own key. A name inherited from `Object.prototype` (a `constructor:` title) passes the command's lookup and crashes it while rendering, with exit 0 and no section written; the script earns nothing for it, so auto-release's rewrite puts the manifest back and its gate skips. Had the script earned a patch there, the gate would have passed on the previous section's entries and shipped a bump, a tag, and a release with empty notes.

Also in this PR: `cli-release.yml` lists `packages/scripts/**` in its push path filter, so a change to the script it runs re-runs the three-platform gate on its own; the scripts page says a hand run needs no network; an auto-release comment that described the old dry run now says the script touches no file; the test fixture gives its throwaway repositories an `origin` remote, since changelogen's config loader reads it to resolve the repository; and the runtime-apis index row for the script drops the import the rewrite removed.

## Measured

In throwaway repositories: the library calls take 27-71ms in process, 70-90ms wall with Bun's startup, where the command took 650-800ms with its author lookup and about 90ms with authors excluded; the same with the network blocked. The answers match the command's for a feature window (0.1.28), a breaking window (0.2.0), `build(deps)` (0.1.28), and a tagless fork's first fix (0.0.1). For a `chore(deps)` plus `ci` window and for a window of one non-conventional commit the script answers nothing earned, where the command bumps the manifest to a patch but renders no entry, which is exactly what the content gate discounts. A broken `changelog.config.json` throws.

## Verification

- 25 tests in `tests/packages/scripts/src/release-version.test.ts`: the pure decision and the bump rule, every stage against real repositories (no tag, feature window, breaking mid-window, mechanical-only, non-conventional and dropped-types windows, a capitalized type, a prototype-key type, already moved, hand-set major, below the tag, unreachable tags), a broken config as a failure, the entry point with and without `--write`, the version field found at the top level, and one run of the real command beside the script on a window with a feature, a dependency bump, and a breaking fix, both answering 0.2.0. That parity scenario is the only place the command still runs, offline through the throwaway config's author exclusion.
- The whole file runs in about 3s where it took about 9s.
- Lint, format, types clean; the dry run on this branch answers 0.1.29 (its own fix commits earn a patch) with the tree untouched.
- The three-platform matrix runs on the canary push after merge, through both `packages/scripts/**` and `tests/**`.

## Summary by CodeRabbit

- **New Features**
  - Release version calculations now run directly through the changelog tooling, including support for prerelease version bump rules.
  - Release workflows now rerun when release scripts change.

- **Bug Fixes**
  - Improved handling of commit types, configuration errors, and release-version calculations without modifying files.

- **Documentation**
  - Clarified that manually running the release-version command only previews the decision and requires no network access.

### DIFF
--- a/.agents/skills/runtime-apis/index.md
+++ b/.agents/skills/runtime-apis/index.md
@@ -31,7 +31,7 @@ runs under the system Node while Docker and Vercel serve it under Bun.
 | `packages/scripts/src/auth-schema.ts` | Bun | `node:path` (join, resolve) |
 | `packages/scripts/src/data-table-metrics.ts` | Bun | `node:path` (join, resolve) |
 | `packages/scripts/src/generate-env.ts` | Bun | `node:path` (resolve) |
-| `packages/scripts/src/release-version.ts` | Bun | `node:path` (dirname, join) |
+| `packages/scripts/src/release-version.ts` | Bun | `node:path` (join) |
 | `tests/github/scripts/ensure-remote-branches.test.ts` | Bun | `node:fs` (mkdtempSync, rmSync); `node:os` (tmpdir); `node:path` (join) |
 | `tests/github/workflows/auto-labeler.test.ts` | Bun | `node:module` (createRequire); `node:path` (join) |
 | `tests/packages/cli/bin/commands/init.test.ts` | Bun | `node:fs` (mkdirSync, mkdtempSync, rmSync, writeFileSync); `node:os` (tmpdir); `node:path` (join) |
--- a/.github/workflows/auto-release.yml
+++ b/.github/workflows/auto-release.yml
@@ -43,7 +43,7 @@ jobs:
           # the boundary this changelog covers; tagged below so the next run starts exactly here
           BOUNDARY=$(git rev-parse HEAD)
           LAST_TAG=$(git describe --tags --abbrev=0 --match "v*" 2>/dev/null || echo "")
-          # the target: the larger of the tree and what the window earned from the last tag (a missing tag counts as v0.0.0); a hand-set ahead of the window stays, one that undercounts it is lifted, a tree below the tag stops here loudly. read before changelogen overwrites package.json below; the script puts every file back after its own dry run
+          # the target: the larger of the tree and what the window earned from the last tag (a missing tag counts as v0.0.0); a hand-set ahead of the window stays, one that undercounts it is lifted, a tree below the tag stops here loudly. read before changelogen overwrites package.json below; the script itself touches no file
           DECISION=$(bun packages/scripts/src/release-version.ts)
           TARGET_VERSION=$(echo "$DECISION" | jq -r .next)
           if [ -z "$TARGET_VERSION" ] || [ "$TARGET_VERSION" = "null" ]; then
--- a/.github/workflows/cli-release.yml
+++ b/.github/workflows/cli-release.yml
@@ -6,7 +6,7 @@ on:
       - canary
     paths:
       - "packages/cli/**"
-      # the release-version script this workflow runs lives here; a change to it re-runs the gate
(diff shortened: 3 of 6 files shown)


nrjdalal:
Browser run on this branch's dev stack (`release-version-offline.api.zerostarter.localhost`), the two surfaces that display the number, before and after the decision moves the tree. The decision came from the library path, no changelogen process, no file touched until `--write`:

```
{"current":"0.1.28","earned":"0.1.29","moved":true,"next":"0.1.29","tag":"v0.1.28"}
```

After `--write` the root `package.json` read 0.1.29, the `@packages/env` build was a cache miss (the `$TURBO_ROOT$/package.json` input from #826), and the restarted API served the new number. The tree was restored afterwards.

| Surface | Before | After |
| --- | --- | --- |
| Scalar docs, header version | `v0.1.28-7516918f` | `v0.1.29-7516918f` |
| `/api/health` | `0.1.28-7516918f` | `0.1.29-7516918f` |

Screenshots attached in order: docs before, health before, docs after, health after.

01-docs-before

02-health-before

03-docs-after

04-health-after

nrjdalal:
Shipped as **v0.1.29**, the first release decided by the library path in this PR. Observed end to end:

| Stage | Result |
| --- | --- |
| Merge push, version step | `{"current":"0.1.28","earned":"0.1.29","moved":true,"next":"0.1.29","tag":"v0.1.28"}`, committed as `ci(version): bump to v0.1.29` (44d848bc). Decision printed 124ms after the install finished. |
| Release Package matrix, canary push | Ubuntu 3.5s, macOS 8.1s, Windows 20.5s, 323 pass / 0 fail each. First run through the new `packages/scripts/**` path filter. |
| Canary prerelease | `zerostarter@0.1.29-canary.0`, version from the same decision. |
| Release merge | #829 into main as 272e7e65, tree already at 0.1.29. |
| auto-release | Tag `v0.1.29` at the boundary 44d848bc, changelog section, GitHub release Latest. |
| CLI publish | All three legs green first try, `zerostarter@0.1.29` on npm latest, `npx zerostarter@0.1.29 --version` answers 0.1.29. |
| Production | web and API both serve `0.1.29-272e7e6`, the released number on the release merge. |

The macOS leg that stalled 30s per decision test on v0.1.28 and needed a re-run finished the whole suite in 8.1s here, on both matrix runs.

Screenshots: production docs and API reference, both showing 0.1.29.

01-prod-docs

02-prod-api-docs

PR #840, opened 2026-09-18
build(deps): refresh the catalog to the latest in-range versions

nrjdalal:
## Summary

Lockfile deleted and rebuilt, then `bun update --recursive` inside the caret ranges the catalog already declares. A day after the last refresh, two entries move and neither crosses a major:

- **turbo 2.10.13 to 2.11.1**, a minor on the tool that runs every build, `check-types` and test task.
- **posthog-js 1.433.7 to 1.434.0**.

Every spec stays a caret range, so `deps-manager.ts` has nothing to rewrite on the next install.

## Verification

The gate, in order, all green:

- `bun audit --audit-level high`: no vulnerabilities across 1074 packages.
- `bun run lint`, `bun run format`: clean, formatter rewrote no files.
- `bun run check-types`: 13 of 13. `bun run test`: 323 pass, 67 skip, 0 fail. `bun run build`: 7 of 7.
- `bunx turbo --version` reports **2.11.1**, so the new binary is the one that drove those runs rather than a cached older one.

posthog-js runs in the browser, so the gate alone is not enough. Checking it turned up something worth stating: the app initialises PostHog in `web/next/instrumentation-client.ts`, and **only when `NEXT_PUBLIC_POSTHOG_PROJECT_TOKEN` is set**, which the example env leaves blank. A default local run therefore never loads the library at all, so a browser pass against the stock env would have proved nothing about this bump.

So the check was run with a token set and `NEXT_PUBLIC_POSTHOG_HOST` pointed at a closed local port, which exercises the new code without sending anything anywhere:

- PostHog initialised: it wrote its own storage key `ph_<token>_posthog`, and made 3 requests, all to the dead local host.
- With analytics live the home page still rendered and a client route change to `/docs` worked, with no errors in the dev log.
- Signed in as the local agent against a disposable Postgres, the dashboard and the console users table render with no horizontal overflow.

Screenshots in the comment below. Rows are the disposable database's own; no real data.

## Note on the title

`build(deps)`, not `chore(deps)`: changelogen drops non-breaking `chore(deps)`, so a window holding only one can never release.

## Summary by CodeRabbit

* **Chores**
  * Updated internal catalog dependencies to newer versions for improved compatibility and maintenance.


nrjdalal:
Browser verification for the posthog-js bump, against a disposable Postgres with analytics pointed at a closed local port.

The app only loads PostHog when a project token is set, so the stock env never initialises it; these were taken with a token set, which is what makes the bump actually exercised. PostHog wrote its storage key and made 3 requests, all to the dead local host, and the pages rendered alongside it with no dev-log errors.

Home, analytics off (stock env): the library never loads

Home with posthog-js 1.434.0 initialised against a dead local host

Dashboard, signed in as the local agent

Console users table, disposable database rows only

PR #842, opened 2026-09-18
build(deps): take the six entries that sit outside their caret ranges

nrjdalal:
## Summary

Six catalog entries sit beyond their caret ranges, so no in-range refresh reaches them. Taken **all at once on purpose**, to find out what breaks rather than guessing one at a time:

| Package | From | To |
| --- | --- | --- |
| `@hono/standard-validator` | 0.3.0 | 0.4.0 |
| `@scalar/hono-api-reference` | 0.11.16 | 0.12.2 |
| `cn` | 0.2.6 | 0.3.0 |
| `dotenv` | 17.4.2 | 18.0.0 |
| `oxfmt` | 0.62.0 | 0.68.0 |
| `tsdown` | 0.22.14 | 0.23.0 |

`dotenv` is a real major; the rest are `0.x` minors, which semver treats the same way. Every spec stays a caret range.

**Nothing broke.** Two results were better than expected: `oxfmt` reformatted **zero files** across six minors, and `dotenv` 18 boots the stack and runs migrations unchanged.

## Verification

Gate, all green: `bun audit` clean, lint and format with **no file rewritten**, `check-types` 13 of 13, unit suite 323 pass / 67 skip / 0 fail, `build` 7 of 7.

Then the parts the gate cannot see, against a local-stage stack on a disposable Postgres:

- **dotenv 18**: migrations ran and `/api/health` answered 200, so env loading is intact across the api, the web app and the scripts.
- **@hono/standard-validator 0.4**: a malformed body still answers `400` with the repo's `{ error: { code, message } }` envelope, and a missing field still answers `VALIDATION_ERROR`.
- **@scalar 0.12**: `/api/docs` serves 200 and `/api/openapi.json` still resolves.
- **tsdown 0.23**: every package and the api bundle build.
- **cn 0.3**: home, dashboard and the console users table render identically to `canary` at the same viewport, with no horizontal overflow, and merged classes still resolve to a single winner. Before and after are in the comment below.

**The end-to-end suite is 55 pass / 2 fail on this branch, and the identical 55 pass / 2 fail on plain `canary`**, so neither failure comes from these bumps. One was my own test environment (blank OAuth client ids hide two providers from a snapshot). The other is a genuine pre-existing break that this exercise uncovered, fixed separately in #841: zod 4.6.5 rewrote the regex captured in the waitlist snapshot. With #841 merged, this branch's suite is 57 pass / 0 fail.

## Suggested order

#841 first, since it fixes a break already shipped in two releases. This one after.


nrjdalal:
`cn` 0.3.0 changes class merging across 73 files, so here is the same surface before and after at 1782x972, signed in as the local agent against a disposable Postgres. The console users table is the densest surface in the app: sidebar, sortable header, badges, footer counts. It renders identically; only the row's date differs, because the two runs are a day apart.

Before: console users on canary, cn 0.2.6

After: the same page with cn 0.3.0

Before: dashboard on canary

After: dashboard with cn 0.3.0

After: the marketing home with all six bumps

